PR #42, opened 2014-08-19
Load error logs

cvicente:
Improves logging of errors generated by cu.load(). Deals with older versions of Junos which incorrectly send the \</ok\> element in the response. 

Sample log:

2014-08-19 17:26:45,947:CONFIG:vsrx:unable to load config: severity:'error', msg:'syntax error', bad element:'blah'


shermdog:
This has been superseded by #49.  Thank you for the report though, this has been on our dirty laundry list for a while!

cvicente:
That's much better. Thank you!

PR #55, opened 2015-02-12
Adds support for the "replace:" statement

XioNoX:
Supported on Junos_PyEZ, this allows configuration files with the "replace:" statement to be pushed and correctly interpreted (similar to a "load replace" on Juniper CLIs.

Please let me know if I need to change anything.

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -204,10 +204,13 @@ def junos_install_config(module, dev):
             replace = module.boolean(module.params['replace'])
             if True == overwrite:
                 load_args['overwrite'] = True
+                load_args['merge'] = False
             elif False == overwrite:
                 load_args['merge'] = True
-            elif False == replace:
+                load_args['overwrite'] = False
+            if True == replace:
                 load_args['merge'] = False
+                load_args['overwrite'] = False
             cu.load(**load_args)
         except ValueError as err:
             logging.error("unable to load config:{0}".format(err.message))
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -84,6 +84,13 @@ options:
         required: false
         default: no
         choices: ['true','false','yes','no']
+    replace:
+        description:
+            - Specify whether the configuration I(file) uses "replace:" statements.
+              (NETCONF only)
+        required: false
+        default: no
+        choices: ['true','false','yes','no']
     timeout:
         description:
             - Extend the NETCONF RPC timeout beyond the default value of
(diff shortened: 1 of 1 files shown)


shermdog:
Hi @XioNoX ,

At current `load replace` is not part of our plans for supported use cases.

With the nature of Ansible and Junos we must do single atomic commits.  At this time we will only be supporting NOOB and Golden Config.

I may leave this open a while for folks to comment or cherry-pick.

Thanks,
-Rick

XioNoX:
Thanks for your quick reply,
Just curious, what's the issue with using replace? I'm only using it with Ansible on SRXs and it seems fine so far.

shermdog:
Management becomes difficult when Ansible is not the authoritative place for configuration.  Further, if users being to load multiple independent changes, Ansible must open a device, lock, load, commit, and unlock every time the module is called in a playbook.  It's a use case that is difficult for us to support as a majority of network professionals are new to automation tools and programming.

The savvy users are generally able to mitigate the pitfalls and as such are generally able to add in the little code hook they need for `replace`.  :)

-Rick

XioNoX:
load replace still let Ansible be the authoritative place for configuration and can be used in 1 single "commit".
The main advantage is to be able to progressively mange the configuration using Ansible instead of having to do "all or nothing".

For example if I already have a working configuration with that tree:
security {
    policies {
        [...]
   }
    ipsec {
        [...]
   }
    nat {
        [...]
   }
}

I can generate a configuration file that looks like:
security {
    replace: ipsec {
    /* Managed by Ansible */
        [...]
   }
}

And only modify the ipsec block without impacting anything else under security.
Then later commit:
security {
    replace: policices {
    /* Managed by Ansible */
        [...]
   }
   replace: ipsec {
    /* Managed by Ansible */
        [...]
   }
}
That way Ansible is authoritative on the sections we care about (and it's clearly defined).
And that only needs 1 commit for the whole tree (from the root).
The replace statement is actually more authoritative than a load merge (I think) as it replace the whole block and removes sub blocks when not in the authoritative config file anymore.
Another advantage is that I can split the configuration into different roles (1 role security policies/addressbook, 1 NAT, 1 users, etc... reuse them between devices of different kinds, and only let some users impact some parts of the config.
Last point, I'm wondering if it would be still possible to add the functionality (even with a disclaimer) so people who use it don't have to maintain the feature on their side.

Thanks :)

cvicente:
Just FYI, we write our templates in XML, and use the "operation" parameter to specify the action per tag, for example:

<snmp operation="replace">
...
</smp>

This allows us to control only parts of the config with Ansible, so we can add  incrementally.

sean-:
We're interested in the replace functionality too, specifically for policies and ACLs that we compile externally and load as a replace.  Ansible doesn't handle the lifecycle of managed assets well unless you explicitly state=absent each item, which becomes klunky and is nearly impossible when using precompiled sections (as would be effectively required when compiling ACLs).

rodecker:
I would like to see this patch merged.

PR #57, opened 2015-02-16
support commit confirm

pamelia:
Hello

I don't expect this to be merged just like that but I'm more interested in opening up for a discussion. JUNOS and commit confirmed is really awesome and so is Ansible and wait_for. Example usage: http://p.ip.fi/OWJg
The commit_check.py script of course uses py-junos-eznc and connects to the device and does a commit_check().

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -126,6 +126,12 @@ options:
             - TCP port number to use when connecting to the device
         required: false
         default: 830
+    confirm:
+        description:
+            - Using commit confirmed, number of minutes until automatic
+              rollback
+        required: false
+        default: None
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


shermdog:
Hi @pamelia ,

It is an interesting topic of using some external scripts to do validations.  You are correct that this will most likely not make it into our code base though.

The current example of what you are doing is unnecessary. 

 * The module loads the configuration in 'exclusive' mode via a lock().
 * If the configuration is unable to be committed a 'CommitError' is thrown
 * Upon exiting the module the NETCONF connection is closed and the config is automatically rolled back.

pamelia:
Hi @shermdog 

Lets say I manage the whole config with ansible, and lets say I change a firewall filter or routing policy. If I use ansible and junos_install_config the commit can be successful and still break things.
I use wait_for after the "commit confirmed" version of junos_install_config to *verify* that I have connectivity to the device I'm managing. If I can reach the device on port 22 I call my script that does a commit check to abort the automatic rollback commit confirmed does. If I can't reach the device the automatic rollback will take place and things are not so broken.

Aaronprat:
Wonderful, exactly what I looked for. Now I don't have to write it

Thanks!

PR #76, opened 2015-08-03
Feature - allow ssh private key

thomseddon:
Hey,

As title says, add support for passing a private key file - Depends on https://github.com/Juniper/py-junos-eznc/pull/418 to be useful :)

### DIFF
--- a/library/junos_commit
+++ b/library/junos_commit
@@ -107,13 +107,14 @@ def main():
 
     module = AnsibleModule(
         argument_spec=dict(host=dict(required=True, default=None),  # host or ipaddr
-                           user=dict(required=False, default=os.getenv('USER')),
-                           passwd=dict(required=False, default=None),
-                           port=dict(required=False, default=830),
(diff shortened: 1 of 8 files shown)


thomseddon:
(removed e13a69af which I will submit separately)

stacywsmith:
This is now supported in common code for all modules via #278.

All modules support the following:
```
  ssh_private_key_file:
    description:
      - The path to the SSH private key file used to authenticate with the
        Junos device. If this option is not specified, and no default value is
        found using the algorithm below, then the SSH private key file
        specified in the user's SSH configuration, or the
        operating-system-specific default is used.
    default: The first defined value from the following list:
             1) The ANSIBLE_NET_SSH_KEYFILE environment variable.
                (used by Ansible Tower)
             2) The value specified using the --private-key or --key-file
                command line argument.
             3) none
    type: path
    aliases:
      - ssh_keyfile
```

PR #77, opened 2015-08-11
Use consistent shebang syntax in all Python modules

SegFaultAX:
Quick fix to update all the shebangs in `library/` to use `/usr/bin/env python` instead of hardcoding to `/usr/bin/python`.

Note: Hardcoding specific Python runtimes like `/usr/bin/python` breaks execution environments like virtualenv that install new Python versions on the `PATH`

### DIFF
--- a/library/junos_get_facts
+++ b/library/junos_get_facts
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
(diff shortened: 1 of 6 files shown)


vnitinv:
@SegFaultAX I am sorry I missed your pull request. I did similar changes last month or so without noticing your PR.
Thanks a lot for your contribution @SegFaultAX

PR #78, opened 2015-08-15
Added the ability to use jinja templating with junos_install_config

lamoni:
Since pyEZ already supports templating with jinja by feeding a dict to the cu.load() call, I just added a new ansible module parameter called "template_vars_file" that gets read in, passed through json.load(), and then put in with the load_args array that gets fed to cu.load().  I don't think using the json module will cause any dependency changes, but in the event it does, we could always fall back to the uglier "eval" solution (and assuming we want to put up with cries of bad security).  Please note that the following example is extremely simplified / contrived as to show the use of this.  There are obviously better use cases.

Examples:
## Playbook (ansible-op-scripts-playbook.yml)

```
lamoni@ubuntu:~/playbooks$ cat ansible-op-scripts-playbook.yml 

---
- name: Load merge a configuration to a device running Junos OS
  hosts: srx1
  roles:
  - Juniper.junos 
  connection: local
  gather_facts: no

  tasks: 
  - name: Checking NETCONF connectivity
    wait_for: host={{ inventory_hostname }} port=830  timeout=5

  - name: load merge a configuration file
    junos_install_config: 
      host={{ inventory_hostname }}
      file=/home/lamoni/playbooks/op_script.set
      template_vars_file=/home/lamoni/playbooks/op_script.json
      overwrite=false
      logfile=/var/log/ansible/junos/config.log
```
## Junos Config File (op_script.set)

```
lamoni@ubuntu:~/playbooks$ cat op_script.set
set system scripts op file {{ scriptName }}
```
## JSON Template Vars File (op_script.json)

```
{"scriptName":"somenewscript.slax"}
```

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -77,6 +77,11 @@ options:
               extension, the content is treated as Junos OS B(set)
               commands.
         required: true
+    template_vars_file:
+        description:
+            - Path to the file containing a JSON string that gets
+              converted to a dict and fed to PyEZ for jinja templating
+        required: false
     overwrite:
         description:
             - Specify whether the configuration I(file) completely replaces
@@ -198,6 +203,7 @@ def _load_via_netconf(module):
     in_check_mode = module.check_mode
     overwrite = module.boolean(module.params['overwrite'])
     replace = module.boolean(module.params['replace'])
+    template_vars_file = module.params['template_vars_file']
 
     if all([overwrite, replace]):
         msg = "Overwrite and Replace cannot both be True!"
@@ -244,7 +250,22 @@ def _load_via_netconf(module):
         # an exception if there is even a warning.
         # so we want to avoid that condition.
         logging.info("loading config")
-        load_args = {'path': file_path}
+        load_args = {}
+
+        # normal config load from a 'static' file
+        if template_vars_file is not None:
+            # template config load from template,
+            # so load the template vars file
+            load_args['template_path'] = file_path
+
+            template_vars_file = os.path.abspath(template_vars_file)
+            logging.info("reading from template vars file: {0}".format(template_vars_file))
+            with open(template_vars_file) as vars_file:
+                load_args['template_vars'] = json.load(vars_file)
+
+        else:
+            load_args['path'] = file_path
+
         if replace is True:
             load_args['merge'] = False
         elif overwrite is True:
@@ -391,6 +412,7 @@ def main():
             passwd=dict(required=False, default=None),
             console=dict(required=False, default=None),
             file=dict(required=True),
+            template_vars_file=dict(required=False, default=None),
             overwrite=dict(required=False, choices=BOOLEANS, default=False),
             replace=dict(required=False, choices=BOOLEANS, default=False),
             logfile=dict(required=False, default=None),


lamoni:
Any update on this?

vnitinv:
Hi @lamoni this really makes sense. but I need some time to test it before merging the same.

lamoni:
@vnitinv Thanks Nitin, I'm currently using my fork to demo capabilities, and would like to be using the official repo instead.

library/junos_install_config
```diff
@@ -198,6 +203,7 @@ def _load_via_netconf(module):
     in_check_mode = module.check_mode
     overwrite = module.boolean(module.params['overwrite'])
     replace = module.boolean(module.params['replace'])
+    template_vars_file = module.params['template_vars_file']
```
vnitinv: @lamoni As you can notice its tried to keep in sync with the variable name being used in PyEZ directly. For ex: 
overwrite = module.boolean(module.params['overwrite'])
replace = module.boolean(module.params['replace'])
Similarly can we have template_vars  in place of template_vars_file (ref https://junos-pyez.readthedocs.org/en/1.2.3/jnpr.junos.utils.html#jnpr.junos.utils.config.Config.load)

RiotGeneral:
Any word on the status of this pull request? Would like to start testing with this feature.

lamoni:
@vnitinv Pertaining to your note, I can definitely make that change.  Do you prefer I rebase the two commits or is this project okay with multiple incremental commits in a pull request?

stacywsmith:
@lamoni I'm trying to understand the benefit of this proposed feature over simply creating a play with a task that calls the current Ansible template core module followed by a task (or handler) which calls junos_install_config to install the newly generated file. Because the core template module only sets 'changed' when the rendered file has actually changed, you can avoid the install/commit if the configuration has not changed. In addition, all of the Ansible variables/facts are available from within the template core module without having to pass any specific arguments as you would have to do with this feature.

Is there some additional benefit this feature would provide that I'm missing?

lamoni:
@stacywsmith I admittedly don't consider myself a poweruser on Ansible, is there a way to feed in the variables used for the jinja template in the template core module that doesn't require "hardcoding" it in the playbook?  My main concern was separation of concerns, as far as the playbook is concerned.  I'm demo'ing the ability to automate MOPs, and am showing they can write the process once, and then never touch that file again (and instead just touch the file where the variables used in the MOP are stored). So in my example in the pull request, they'll only have to ever touch the .json file, and that is easily generated by a script (rather than generating the entire playbook file or splicing in the "vars" into an existing playbook file).  When I first added this code, I did a quick skim and it seemed I could only pass the variables in through the playbooks, so I definitely could have (if not probably) overlooked that.

stacywsmith:
@lmoni Yes, the variables can be set in any of the myriad of places that Ansible allows setting variables. This includes inventory variables, group variables, via "register" from previous tasks, in the vars directory of a role, etc.

There's a good example of this approach at: https://github.com/dgjnpr/ansible-template-for-junos

stacywsmith:
This has now been implemented in the new `juniper_junos_config` module introduced via #290.

This module now supports the `template` and `vars` option. Example playbook:
```
    - name: Load from a Jinja2 template.
      juniper_junos_config:
        load: 'merge'
        format: 'xml'
        template: "{{ inventory_hostname }}.j2"
        vars:
      register: response
    - name: Print the complete response.
      debug:
        var: response
```

PR #89, opened 2015-10-28
Inventory

spidercensus:
Hey guys. I created a script that imports the devices from a Space instance using the REST API, and then formats those into hosts for use in an Ansible inventory.

Attributes of the devices including platform, os version, ip, and hostname are added to the inventory as host metavars.


spidercensus:
Here are the specs for an inventory script.  Sorry for the mess, this is my first pull request. 

http://docs.ansible.com/ansible/developing_inventory.html

PR #99, opened 2016-01-22
Space Inventory script

spidercensus:
Back by popular demand.  

This is a slight rewrite of my original script which outputs a Space inventory into an Ansible-digestible format.


vnitinv:
Hi Jason @spidercensus. As this library goes as Juniper.junos Roles, shouldn't we keep this space module separate as Juniper.space Role?

tools/inventory.py
```diff
@@ -0,0 +1,200 @@
...
+  return config;
+  
+if __name__ == "__main__":
+  args = parseArgs()
+  config = loadConfig(args.config_file)
+  [host, user, password] = [ config['SPACE_HOST'], config['SPACE_USER'], config['SPACE_PASSWORD'] ]
+  space = Space(host, user, password)
+  space.getDevices()
+  data = space.exportJson()
+  if args.opt_hostj and args.opt_host in data['_meta']['hostvars'].keys():
```
MillsyBot: Please change args.opt_hostj to args.opt_host

tools/inventory.py
```diff
@@ -0,0 +1,200 @@
...
+    if i not in config.keys():
+      raise Exception("Config file %s is missing config value %s.\n" % (filename, i)) 
+  return config;
+  
+if __name__ == "__main__":
+  args = parseArgs()
+  config = loadConfig(args.config_file)
+  [host, user, password] = [ config['SPACE_HOST'], config['SPACE_USER'], config['SPACE_PASSWORD'] ]
+  space = Space(host, user, password)
+  space.getDevices()
```
MillsyBot: Line 194 is not needed as it is called in line 138 devices = self.getDevices() by line 195 data = space.exportJson().

tools/inventory.py
```diff
@@ -0,0 +1,200 @@
...
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#================================#
+#====== Lawyer stuffs above.=====#
+#================================#
+
+import optparse
+import requests
+import json
+import sys
+import os
```
MillsyBot: from requests.packages.urllib3.exceptions import InsecureRequestWarning 
This needs to be added to the import to silence in secure requests. Ansible doesn't like the the extra return values.

tools/inventory.py
```diff
@@ -0,0 +1,200 @@
...
+
+  #validate config
+  necessary = [ 'SPACE_HOST','SPACE_USER','SPACE_PASSWORD']
+  for i in necessary:
+    if i not in config.keys():
+      raise Exception("Config file %s is missing config value %s.\n" % (filename, i)) 
+  return config;
+  
+if __name__ == "__main__":
+  args = parseArgs()
```
MillsyBot: requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
This line is needed to silence the insecure warnings from Space. Ansible doesn't like it.

stacywsmith:
This needs to be published under some other repository as none of the other functionality uses or depends on Space.

PR #100, opened 2016-01-28
Enhancement/config diff

chepazzo:
I added a few things:

1. junos_get_config now supports check_mode (as opposed to claiming that it did, but overwriting the `dest` file anyway).
2. Properly reports 'changed' attribute for junos_get_config module.
3. Fixed the requirement status for the dest argument in junos_get_config (docs claimed True, but set to False when module was called).
4. If `--diff` on cmdline*, return diff diff with results.
   - In junos_get_config, sets `before` and `after` attributes and lets Ansible do the diff.
   - In junos_install_config, sets `prepared` with the output of `commit check`.

\* Currently, Ansible 2.0.0.2 does not report this, but the devel branch shows setting `module._diff` as well as reading `diff['prepared']` from returned results. While this pull request will ignore `--diff` cmdline args in 2.0.0.2, it will accept and properly handle that option when Ansible upgrades.

### DIFF
--- a/library/junos_get_config
+++ b/library/junos_get_config
@@ -143,6 +143,7 @@ def main():
         module.fail_json(msg='junos-eznc >= 1.2.2 is required for this module')
 
     args = module.params
+    results = {}
 
     logfile = args['logfile']
     if logfile is not None:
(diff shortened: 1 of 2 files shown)


vnitinv:
@sdndude @stacywsmith @dgarros  Can you please help me if this feature need to go into the module itself. Should we be doing this comparison using different tasks. We can make destination file dynamic and compare them using different modules (core)
What can be the best approach to handle such feature?

dgarros:
Hi @vnitinv, this pull request looks good to me, lots of good fixes. 
Thanks a lot @chepazzo

Issue #102, opened 2016-02-05
Traceback TypeError with Ansible 2.0.0.2


jamesboswell:
Juniper.junos breaks with Ansible 2.0.0.2

Ansible:  2.0.0.2 from Hombrew (http://brew.sh) build: (https://github.com/Homebrew/homebrew/commit/81ec8e651d7d7b84b2e36717bc6294eebeb98f2b)
junos-eznc==1.3.1 (installed directly from pip)

**Error on ansible 2.0.0.2**

```python
An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 2143, in <module>
    main()
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 207, in main
    module.exit_json(**m_results)
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 1743, in exit_json
    kwargs = remove_values(kwargs, self.no_log_values)
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 626, in remove_values
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 626, in <genexpr>
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 626, in remove_values
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 626, in <genexpr>
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/Users/test/.ansible/tmp/ansible-tmp-1454709965.92-125910454178634/junos_get_facts", line 635, in remove_values
    raise TypeError('Value of unknown type: %s, %s' % (type(value), value))
TypeError: Value of unknown type: <class 'jnpr.junos.facts.swver.version_info'>, junos.version_info(major=(11, 4), type=R, minor=5, build=7)

[...]

PR #103, opened 2016-02-10
Addressing issue #102.

stacywsmith:
Ansible 2 adds a feature to search the modules return value for any
strings which should not be logged. This feature only handles built-in
data types and does not handle custom objects.

The junos_get_facts module was returning a junos.version_info object
which caused this new feature to raise a TypeError.

Since all Ansible return values are passed via JSON, the junos.version_info
object was previously getting converted to a dictionary anyway, so this
change should be backwards compatible with Ansible 1 as well.


vnitinv:
Thanks @stacywsmith

PR #104, opened 2016-02-10
Add role junos_cli to execute CLI on device and save output locally

dgarros:
Hi
I created a new role to simply execute a show /request cli command on a device and save the result locally
As discussed, it will be great to convert XML to JSON properly and feed JSON natively into Ansible, this part is not available today 

Thanks
Damien

### DIFF
--- a/library/junos_cli
+++ b/library/junos_cli
@@ -1,0 +1,183 @@
+#!/usr/bin/env python
+
+# Copyright (c) 1999-2015, Juniper Networks Inc.
+#               2014, Jeremy Schulman
+#               2015, Rick Sherman
+#
+# All rights reserved.
+#
+# License: Apache 2.0
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright
+#   notice, this list of conditions and the following disclaimer.
+#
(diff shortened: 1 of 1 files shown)


library/junos_cli
```diff
@@ -0,0 +1,183 @@
...
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+DOCUMENTATION = '''
+---
+module: junos_cli
+author: Damien Garros, Juniper Networks
+version_added: "1.2.0"
+short_description: Execute CLI/RPC on device and save the output locally
```
vnitinv: @dgarros  Why are you mentioning rpc. I dont see this module executing RPCs

library/junos_cli
```diff
@@ -0,0 +1,183 @@
...
+description:
+    - Execute CLI/RPC on device and save the output locally on a file
```
vnitinv: same as above. 

I had written a module to run RPCs (to take args too). Will share you the same.

PR #105, opened 2016-02-12
To call any junos rpc.

vnitinv:
My playbook for the same

```yaml
---
- name: Junos OS version
  hosts: all
  connection: local
  gather_facts: no
  tasks:
    - name: Get rpc run
      junos_rpc:
        host={{ inventory_hostname }}
        user=xxxx
        passwd=xxx
        rpc=get-interface-information
        kwargs="interface_name=em0"
        format=text
        dest=get_interface_information.conf
      register: junos

    - name: version
      debug: msg="{{ junos.rpc_reply }}"
```

### DIFF
--- a/library/junos_rpc
+++ b/library/junos_rpc
@@ -34,7 +34,7 @@
 
 DOCUMENTATION = '''
 ---
-module: junos_run_rpc
+module: junos_rpc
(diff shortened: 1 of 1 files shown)


fooelisa:
for what it's worth, i just tested this and it seems to be working well!

PR #106, opened 2016-02-12
adding config mode=private in junos_install_config

fooelisa:
Added ability to init config object in private mode for junos_install_config. I guess dynamic/exclusive/batch could be added in a similar fashion.

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -99,6 +99,12 @@ options:
               timeout interval.
         required: false
         default: "0"
+    mode:
+        description:
+            - Set mode to 'private' to work in private database
+        required: false
+        default: None
+        choices: ['private']
     logfile:
         description:
             - Path on the local server where the progress status is logged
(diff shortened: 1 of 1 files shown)


vnitinv:
@fooelisa you made the changes just to add private behavior. Can I cherry pick your changes and add more pieces to supported all other modes?

fooelisa:
yes please! I haven't used the other modes though pyez, so I wasn't sure if they work the same way, feel free to add them all

stacywsmith:
This is now supported in the new `juniper_junos_config` module introduced via #290.

Example:
```
    - name: Append .foo to the hostname using private config mode.
      juniper_junos_config:
        config_mode: 'private'
        load: 'merge'
        lines:
          - "set system host-name {{ inventory_hostname }}.foo"
      register: response
    - name: Print the config changes.
      debug:
        var: response.diff_lines
```

PR #117, opened 2016-02-29
Regular expression changes to handle lot more key value arguments

vnitinv:


vnitinv:
@dgarros Can you review this pull request, so that it can be merged.

Issue #124, opened 2016-03-29
Issue with junos_get_table and RouteSummaryTable


XioNoX:
Trying to get the routing table using the following playbook:
```
- name: Get structured routing table
  hosts: all
  connection: local
  gather_facts: False
  roles:
    - Juniper.junos
  tasks:
    - name: Get route table
      junos_get_table:
        table=RouteSummaryTable
        file=routes.yml
        host={{ inventory_hostname }}
        port=22
      register: result

    - debug: var=result
```

I get the following error:
```

An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 2198, in <module>
    main()
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 235, in main
    module.exit_json(resource=resource)
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 1776, in exit_json
    kwargs = remove_values(kwargs, self.no_log_values)
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 655, in remove_values
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 655, in <genexpr>
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 653, in remove_values
    return [remove_values(elem, no_log_strings) for elem in value]
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 655, in remove_values
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 655, in <genexpr>
    return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
  File "/home/xionox/.ansible/tmp/ansible-tmp-1459256922.12-4199303295019/junos_get_table", line 666, in remove_values
[...]

PR #127, opened 2016-04-07
fix for nested tables

jedelman8:
fix for issue #124


stacywsmith:
FWIW, this fix looks good to me.

jedelman8:
Don't merge yet, need to add something.

jedelman8:
Okay, should be good to go.

PR #128, opened 2016-04-08
Add ping module

dgarros:
Hi 

I created a new module to execute ping on device
I thought about reusing the rpc module but I thought this one is very specific and there's a value to have it as a standalone module

It's very useful to test L3 IP fabric connectivity

Damien


dgarros:
@vnitinv @stacywsmith 
Please can you take a look and let me know if it looks good to you
Thanks
Damien

vnitinv:
@dgarros Looks good to me. Just one feedback. Looks like you missed including rapid description in docstring. Can you please add that

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+    required: false
+  ttl:
+    description:
+        - Number of device allowed between source and destination
+    required: false
+  routing_instance:
+    description:
+        - Name of the routing instance to use to send the ping
+    required: false
+  source_ip:
```
stacywsmith: source_ip argument is included twice.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+    description:
+        - Destination ip
+    required: true
+  source_ip:
+    description:
+        - Source IP used to send the ping
+    required: false
+  ttl:
+    description:
+        - Number of device allowed between source and destination
```
stacywsmith: Suggest changing to:
"Maximum number of IP routers (IP hops) allowed between source and destination"

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        argument_spec=dict(
+            host=dict(required=True),
+            logfile=dict(required=False, default=None),
+            user=dict(required=False, default=os.getenv('USER')),
+            passwd=dict(required=False, default=None),
+            port=dict(required=False, default=830),
+            dest_ip=dict(required=True, default=None),
+            source_ip=dict(required=False, default=None),
+            interface=dict(required=False, default=None),
+            rapid=dict(required=False, default=True),
```
stacywsmith: As Nitin said, this option is not documented.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        HAS_PYEZ = True
+except ImportError:
+    HAS_PYEZ = False
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            host=dict(required=True),
+            logfile=dict(required=False, default=None),
```
stacywsmith: option is not documented.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+            error_message = rpc_reply.find('rpc-error/error-message').text
+            if error_message is not None:
+                msg = error_message.strip()
+
+            logging.error(msg)
+            dev.close()
+            module.fail_json(msg=msg)
+
+        # Extract packet loss
+        packet_loss = rpc_reply.xpath("//ping-results/probe-results-summary/packet-loss")[0].text.strip()
```
stacywsmith: This line will raise an IndexError exception of the XPath doesn't match a node. Instead, I suggest this expression which returns an empty string if the XPath doesn't mach a node:
packet_loss = rpc_reply.findtext("//ping-results/probe-results-summary/packet-loss").strip()

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        packet_sent = rpc_reply.xpath("//ping-results/probe-results-summary/probes-sent")[0].text.strip()
```
stacywsmith: This line will raise an IndexError exception of the XPath doesn't match a node. Instead, I suggest this expression which returns an empty string if the XPath doesn't mach a node:
packet_sent = rpc_reply.findtext("//ping-results/probe-results-summary/probes-sent").strip()

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        responses_received = rpc_reply.xpath("//ping-results/probe-results-summary/responses-received")[0].text.strip()
```
stacywsmith: This line will raise an IndexError exception of the XPath doesn't match a node. Instead, I suggest this expression which returns an empty string if the XPath doesn't mach a node:
responses_received = rpc_reply.findtext("//ping-results/probe-results-summary/responses-received").strip()

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+
```
stacywsmith: Add a check for packet_loss being an empty string, and set a descriptive error message.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        rpc_reply = dev.rpc.ping(**ping_params)
+
+        # results['xml_reply'] = etree.tostring(rpc_reply)
+
+        # Try to Find probe summary
+        probe_summary = rpc_reply.find('probe-results-summary')
+        if probe_summary is None:
+            results['failed'] = True
+
+            error_message = rpc_reply.find('rpc-error/error-message').text
```
stacywsmith: Can be simplified to:
error_message = rpc_reply.findtext('rpc-error/error-message')

library/junos_ping
stacywsmith: Suggest changing this variable name from packet_sent to packets_sent

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        if int(packet_loss) != 0:
+            results['failed'] = True
+            msg = 'Loss {0}% packets, (Sent {1} | Received {2})'.format(str(packet_loss), str(packet_sent), str(responses_received))
+            logging.error(msg)
+            dev.close()
+            module.fail_json(msg=msg)
+
+    except Exception as err:
+        results['failed'] = True
+        logging.error("unable to execute ping due to:{0}".format(err.message))
```
stacywsmith: Not absolutely required, but suggest a ```dev.close()``` after this line.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+            module.fail_json(msg=msg)
+
+        # Extract packet loss
+        packet_loss = rpc_reply.xpath("//ping-results/probe-results-summary/packet-loss")[0].text.strip()
+        packet_sent = rpc_reply.xpath("//ping-results/probe-results-summary/probes-sent")[0].text.strip()
+        responses_received = rpc_reply.xpath("//ping-results/probe-results-summary/responses-received")[0].text.strip()
+
+        if int(packet_loss) != 0:
+            results['failed'] = True
+            msg = 'Loss {0}% packets, (Sent {1} | Received {2})'.format(str(packet_loss), str(packet_sent), str(responses_received))
```
stacywsmith: Can you set the message before this if block? That way the same message is present on both success and failure.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+            logging.error(msg)
+            dev.close()
+            module.fail_json(msg=msg)
```
stacywsmith: Rather than:
```
module.fail_json(msg=msg)
```
Assuming the msg is set in results['msg'] before the if block, can you:
```
results['failed'] = True
module.exit_json(**results)
```
This allows all of the caller to see all of the parameters specified, the results values, and the msg.

library/junos_ping
```diff
@@ -0,0 +1,248 @@
...
+        if probe_summary is None:
+            results['failed'] = True
+
+            error_message = rpc_reply.find('rpc-error/error-message').text
+            if error_message is not None:
+                msg = error_message.strip()
+
+            logging.error(msg)
+            dev.close()
+            module.fail_json(msg=msg)
```
stacywsmith: If error_message is None, this will raise an exception because msg has not been assigned.

stacywsmith: Suggest changing to:
```
results['msg']=msg
results['failed'] = True
module.exit_json(**results)
```

This allows caller to see the error message, plus the parsed parameters stored in results.

stacywsmith:
@dgarros A couple of general thoughts on this module:

1) You sometimes see a single packet lost due to the time required for ARP resolution. Other times, some level of packet loss might be "acceptable". The module currently only succeeds if there is zero packet loss. Should you offer an ``acceptable_packet_loss`` parameter to allow the user to specify when this module reports success? The ``acceptable_packet_loss`` parameter could default to a value of 0 to maintain the current behavior as the default.

2) This module will always fail if the rapid option is not specified and a count > 30 is specified. That's because probes are sent at 1 second intervals and the default RPC timeout is 30 seconds. I see a couple of potential "solutions": a) Don't allow the user to specify count >= 30 without rapid also being set. b) Allow the user to specify an argument to increase the RPC timeout. c) When count is > 30 and rapid is not set, calculate the RPC timeout as something like int(count * 1.2). This would allow the ping time to complete. In this case, you still might want to have a check to ensure count is within some reasonable limit.

dgarros:
@stacywsmith @vnitinv Thanks for your feedback, 

I've updated the module and I incorporated all suggested changes
I also remove all logging and add an option for timeout and acceptable_packet_loss
should I merge it ?

cvicente:
I'm very interested in using this. Thank you.

library/junos_ping
```diff
@@ -0,0 +1,271 @@
...
+        - Maximum number of IP routers (IP hops) allowed between source and destination
+    required: false
+  routing_instance:
+    description:
+        - Name of the routing instance to use to send the ping
+    required: false
+  timeout:
+    description:
+        - Extend the NETCONF RPC timeout beyond the default value of
+          30 seconds. Set this value to accommodate rollbacks
```
stacywsmith: Copy and paste issue. Please change to: "Set this value to accommodate pings"

library/junos_ping
```diff
@@ -0,0 +1,271 @@
...
+            results['failed'] = True
+            results['msg'] = rpc_reply.findtext('rpc-error/error-message').strip()
+            dev.close()
+            module.exit_json(**results)
+
+        # Extract packet loss
+        packet_loss = rpc_reply.findtext("ping-results/probe-results-summary/packet-loss")
+        packets_sent = rpc_reply.findtext("ping-results/probe-results-summary/probes-sent")
+        responses_received = rpc_reply.findtext("ping-results/probe-results-summary/responses-received")
+
```
stacywsmith: Suggest adding the following lines so the individual return values are also easily accessible in the response, rather than having to parse them from results['msg']:

results['packet_loss'] = packet_loss
results['packets_sent'] = packets_sent
results['packets_received'] = packets_received

stacywsmith:
@dgarros 
I added two minor comments on the latest iteration. Once those are addressed, I'm good with merging.

dgarros:
Thanks @stacywsmith, very good feedback. 
I incorporated your suggestions

PR #131, opened 2016-04-19
Add timeout parameter to junos_rollback

cvicente:


dgarros:
Thanks for the patch, it's missing on many modules.

Issue #122, opened 2016-03-09
junos_cli timeout


Knockturn:
Hi,
I had a problem with timeout while doing a "request" using the junos_cli module.

Please see attached stackoverflow for how I got around it. Perhaps you can add the timeout to this as well in the future?

http://stackoverflow.com/questions/35863701/ansible-juniper-cli-commands-timeout-error/35887731#35887731

vnitinv:
thanks @Knockturn Will try to add timeout to this module.

dgarros:
Hi @Knockturn 
I added a timeout option for both junos_cli and junos_rpc
Please can you check and confirm if it's working good
https://github.com/Juniper/ansible-junos-stdlib/pull/133
Thanks
Damien

jbain:
junos_cli has a typo in it, submitted PR #135

dgarros:
Closing as #135 as been merged

PR #133, opened 2016-04-21
Add Timeout option for junos_cli and junos_rpc (#122)

dgarros:
@vnitinv / @stacywsmith 
I used the same code as in junos_ping and junos_rollback etc ..
it should be straight forward but just in case, please can you take a look before merge ? 

Thanks
Damien


stacywsmith:
Looks good to me.

PR #135, opened 2016-04-29
Fix typo in timeout in junos_cli, simplified default timeout handling

jbain:
there was a typo in junos_cli, module arguments in that file are in 'args' but the code for timeout was copy/pasted form junos_rpc where they are in 'm_args'

The rest is just some cleanup around handling the default value and casting types.

### DIFF
--- a/library/junos_cli
+++ b/library/junos_cli
@@ -68,11 +68,11 @@ options:
         default: 830
     timeout:
         description:
-            - Extend the NETCONF RPC timeout beyond the default value of
-              30 seconds. Set this value to accommodate Cli commands
+            - Set the NETCONF RPC timeout (default value of
+              30 seconds). Set this value to accommodate Cli commands
               that might take longer than the default timeout interval.
         required: false
-        default: "0"
+        default: "30"
     logfile:
         description:
             - Path on the local server where the progress status is logged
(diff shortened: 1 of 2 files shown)


stacywsmith:
I agree we should fix the typo of m_args when it should be args.

I agree with the rewording of the comments.

However, I'm not sure I agree with the "cleanup" in handling the default value. The previous code intentionally uses the PyEZ default value for the RPC timeout (which just currently happens to be 30 seconds). The proposed change hardcodes the default to 30 seconds even if the PyEZ value changes in the future.

I see merit in both approaches, and I see drawbacks in both approaches. If a majority of others agree the approach in this pull request is preferred, I'll go along with the group, but I personally prefer the current behavior.

I just want it to be a conscious decision with consensus.

jbain:
Reasonable. I think my change was a combination of "Make the code match the docs" and "make the docs match the code"

I took issue with the docs specifying "default value of 30" followed immediately by "default: 0", I found it confusing and misleading.

I'd be fine punting to PyEz for the default if that's the preferred way, but would like the docs updated to something like:
```
Set the NETCONF RPC timeout. Adjust this value to accommodate Cli commands that might take longer than the default timeout interval. Setting to 0 uses the PyEZ default (30 seconds)
default: 0
```

stacywsmith:
I totally agree with changing the docs to make it more clear that we're using the PyEZ timeout!

vnitinv:
@stacywsmith @jbain  Code looks good to me now.

jbain:
just wanting to poke at this again. until this (or another PR) is committed to fix the initial m_args -> args typo, junos_cli does not work and ansible fails with a python error.

dgarros:
Thanks for the fix, I'm a bit ashamed of this one, I should know better

PR #136, opened 2016-04-29
Added 'set_facts_as' option to get_junos_facts to register 'facts'

jbain:
using `set_facts_as: key` in a task will copy the returned 'facts' into `ansible_facts.key` so that they will be automatically registered by ansible. This is different from just simply doing `register: key` because it will ONLY register the facts and not the entire output which contains unnecessary information (arguments, password, etc) . It also preserves backwards compatibility with existing scripts by not changing existing behavior.

```
- name: Gather Juniper Facts
  junos_get_facts:
    passwd: "{{ passwd }}"
    set_facts_as: vendor_facts
- name: Debug Vendor Facts
  debug: var=vendor_facts
```

### DIFF
--- a/library/junos_get_facts
+++ b/library/junos_get_facts
@@ -74,6 +74,12 @@ options:
               I(savedir/hostname-facts.json)
         required: false
         default: $CWD
+    set_facts_as:
+        description:
+            - include ansible_facts dictionary key in output and copy 'facts'
+              key to ansible_facts.<provided_key>
+        required: false
+        default: None
     logfile:
         description:
             - Path on the local server where the progress status is logged
@@ -124,7 +130,8 @@ def main():
             savedir=dict(required=False, default=None),
             user=dict(required=False, default=os.getenv('USER')),
             passwd=dict(required=False, default=None),
-            port=dict(required=False, default=830)),
+            port=dict(required=False, default=830),
+            set_facts_as=dict(required=False, default=None)),
         supports_check_mode=True)
 
     m_args = module.params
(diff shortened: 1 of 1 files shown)


jbain:
any thoughts on this?

dgarros:
Hi @jbain 
Personally, I'm not sure to understand the benefit of this new option 
Why not just registering to everything and filter when printing debug ?
```yaml
- name: Gather Juniper Facts
  junos_get_facts:
    passwd: "{{ passwd }}"
  register: vendor_facts
- name: Debug Vendor Facts
  debug: var=vendor_facts.facts
```
the output would be the same as far as I understand
am i missing something ? 
thanks
Damien

jbain:
Hi @dgarros,

Thanks for taking a look! Aside from unnecessary clear-text passwords hanging around, another issue with the current method is it incompatible with fact-caching.

http://docs.ansible.com/ansible/playbooks_variables.html#fact-caching

As far as I can tell, the only way to get facts from a module to cache is to return them under the ‘ansible_facts’ key. Simply doing a ‘register’ or even using the ‘set_fact’ module will not cause the variables to be cached:

http://docs.ansible.com/ansible/set_fact_module.html
```
These variables will survive between plays during an Ansible run, 
but will not be saved across executions even if you use a fact cache.
```

if ‘facts’ were copied into ‘ansible_facts.vendor_facts’ in the return object, it would fix this. (And they would be accessible as vendor_facts.factname, rather than vendor_facts.facts.factname). 

Personally, I like having the option of specifying the key to use, hence the ‘set_facts_as’ option.

Cheers,
Jeremy

dgarros:
Looks good to me
@vnitinv @stacywsmith any comment before merging this one ?

jNicker:
Why add it as another argument?
Why not always register those facts as part of the module?

stacywsmith:
Addressed by #278.

PR #139, opened 2016-05-20
Added 'commit' option to junos_install_config to compare a candidate configuration without committing the change

bryanyabczanka:
I added a commit option to junos_install_config to allow a candidate configuration to be loaded and validated, but not committed. The log generated when junos_install_config is called is conveniently in in patch format, so by foregoing a commit, a full diff of changes is generated, and that can either be reviewed, or added piecemeal with `load patch terminal`.

The default behavior of junons_install_config is not changed because 'commit' defaults to true. An example task to bypass the commit is as follows:

```
- name: Grabbing diff
    junos_install_config:
      host={{ inventory_hostname }}
      user={{ username }}
      passwd={{ password }}
      timeout=300
      file={{ junos_config }}
      diffs_file={{ config_diff_file }}
      overwrite=true
      commit=false
      logfile={{ config_diff_log_file }}
      comment="Configured By Ansible"
      confirm=10
```

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -142,7 +142,7 @@ options:
         description:
             - Commits the configuration after it has been loaded
         required: false
-        default: True
+        default: true
         choices: ['true','false','yes','no']
 '''
 
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -138,6 +138,12 @@ options:
             - Provide a confirm in minutes to the commit of the configuration
         required: false
         default: None
+    commit:
+        description:
+            - Commits the configuration after it has been loaded
+        required: false
+        default: True
+        choices: ['true','false','yes','no']
 '''
 
 EXAMPLES = '''
@@ -198,6 +204,7 @@ def _load_via_netconf(module):
     in_check_mode = module.check_mode
     overwrite = module.boolean(module.params['overwrite'])
     replace = module.boolean(module.params['replace'])
+    commit = module.boolean(module.params['commit'])
 
     if all([overwrite, replace]):
         msg = "Overwrite and Replace cannot both be True!"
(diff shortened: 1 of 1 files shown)


vnitinv:
@stacywsmith @dgarros Can you please review and let me know if we should merge this?

dgarros:
I like the idea of the patch but looking at the code it looks like we are doing a commit check only if commit = True. 
I think it will be great to also do a commit check to properly validate the configuration.

did I missed something ?

stacywsmith:
@vnitinv and @bryanyabczanka 

I like the general idea of being able to check a config and get a config diff. However, I've got a few concerns. These shouldn't necessarily stop us from committing this functionality, but I think we should come to consensus on them before we do.

1) This functionality conflicts with the name of the module, junos_install_config, because the config is no longer installed. We could either consider putting this in a completely different module, or we could decide we don't mind the ambiguity between the module name and module functionality.

2) I agree with @dgarros comment that we should still perform a `commit check` even when `not commit`.

3) If the config is modified, then `results['changed'] = True` is set. In the case of `commit = False`, I don't think we should set `results['changed'] = True`.

4) With this diff, the `commit = False` option is silently ignored if executed via the console (with netconify) and the config would be committed anyway! It may not be easy/possible to implement the `commit = False` functionality with current netconify, but we should at least check for the `not commit and console` condition and fail in that case to avoid an unexpected commit.

5) The module's documentation should be updated to show an example of the `commit = False` case.

6) I have a general concern with how this feature will work within the current Ansible environment.  Because each task (execution of a module) is a new/unique NETCONF connection, there's no guarantee the diff produced by executing `junos_install_config` with `commit = False` is the same diff that would later be produced if the same configuration was loaded and committed. That's because the committed configuration might have changed between the two tasks. This behavior MIGHT be OK, but I'm concerned user's might find it unexpected and confusing.

bryanyabczanka:
@vnitinv, @dgarros, and @stacywsmith:

Regarding stacywsmith's points 2, 3, and 5, I totally agree. The way I implemented the functionality seems a bit lazy, and it would need cleaned up.

Good catch about the silent ignore of commit = False when using console. I'll take another look to see what can be done about that.

As far as the module name junos_install_config, I think it's not a huge deal given that the default behavior would still be to commit the config, but it probably would be best to split off the functionality into a new module--that would solve the naming conflict as well as make the other items easier to address without introducing complexity to the existing module and flow.

For point 6, I think I understand what you're saying; If the junos_install_config module is executed to grab a diff, and exits, and then is subsequently executed in order to commit, someone could have committed another change in the interim. Correct? To me, that's solved by a general workflow process mandated for making changes to devices. I'm not sure what would be done about that as a workaround--I'd think there'd need to be a way to lock and resume the lock on the configuration between sessions, and I'm not aware of a graceful way to do that.

Based on the changes that would need to be made, I think it would be best to create new module (junos_diff_config). If that seems reasonable I would be happy to work on that.

stacywsmith:
This is now implemented in the new `juniper_junos_config` module via #290.

An example is:
```
    - name: Print diff between and commit check, but don't commit
      juniper_junos_config:
        load: 'merge'
        src: "{{ inventory_hostname }}.conf"
        diff: true  # The default
        check: true  # The default 
        commit: false # Not the default. Must be set.
      register: response
    - name: Print the response. Includes the diff.
      debug:
        var: response
```

Issue #145, opened 2016-07-06
junos_get_config doesn't work when using multiple entries in the filter


jedelman8:
When filter is a single element, it works fine.  When I use `filter='interfaces/protocols'`  it no longer works.  Knowing this used to work, I just rolled back to the version before the last commit on this file as shown here:

https://github.com/Juniper/ansible-junos-stdlib/commit/7bfabf0084d335326a93c967d78be199655397bb

And it was fine.  Just an FYI.  Not sure of what that commit was fixing, but may be worth taking another look or let me know where I'm going wrong on filtering multiple config stanzas :)

stacywsmith:
@jedelman8 

> When I use `filter='interfaces/protocols'` it no longer works.

Were you expecting this to provide you with both the `[edit interfaces]` and `[edit protocols]` top-level configuration hierarchies? I believe that's how it used to work

The problem was that there is no way to specify configuration hierarchies with more than one level. For example, if you wanted just the `[edit system login]` hierarchy.

The intention of the `/` syntax in the value of the `filter` argument was to support the latter, not the former. You should be able to retrieve just the `[edit system login]` hierarchy with  `filter='system/login'` and that's what 7bfabf0 fixes.

Now that I see what you are trying to accomplish, I think this could be supported by allowing the value of the `filter` parameter to be a list. I can work on that enhancement when I return from vacation in a couple of weeks.

jedelman8:
Okay, so maybe I happen to use it in the wrong way initially and it just worked, which is fine.  Totally fine either way though.  As long as I know how it _should_ work :)

stacywsmith:
@jedelman8 I agree it was poorly documented. I also agree that both ways _should_ work. I'll work on the enhancement _and_ improving the documentation. Unfortunately, I won't be able to get to it for a couple of weeks. If you need it sooner feel free to submit your own pull request.

vnitinv:
@jedelman8 I am assuming when things were broken for you, you didn't install Junos roles from galaxy but from github.
If I am correct the changes for this issue is not yet published in galaxy.

vnitinv:
@stacywsmith @jedelman8 I need to release github code to galaxy as there are lot many changes gone in from the last release. Can anyone of you do the changes as per discussed above and send us the pull request?

stacywsmith:
This is now possible with the new `juniper_junos_config` module and a `filter_xml` option specified in XML. Please see my comments on #148 for an example.

Issue #120, opened 2016-03-08
junos_get_config module fails when filter argument contains more than one level of configuration hierarchy


stacywsmith:
Per the documentation, the filter argument to junos_get_config:

"Defines heirarchy of configuration to retrieve. If omitted the entire configuration is retrieved. Format is slash notation ex groups/routeinst/routing-instances/ISP-1"

However, the filter argument was not working correctly if it contained more than one level of configuration hierarchy (a string with one or more slashes) per this email report from Dan Houtz:

```
Hi,

I am running into an issue with the following play which makes use of junos_get_config and it's filter parameter:

$ cat mgmt-pfxlist-report.pb.yml
---
- name: MANAGEMENT Prefix List Report
  hosts: all
  connection: local
  gather_facts: no

  vars_prompt:
    - name: USERNAME
      prompt: User name
      private: no
    - name: DEVICE_PASSWORD
      prompt: Device password
      private: yes

  tasks:
    - name: Get MANAGEMENT prefix list
      junos_get_config:
        host={{ansible_host}}
        user={{USERNAME}}
        passwd={{DEVICE_PASSWORD}}
        logfile=/tmp/get_config.log
        filter="system/syslog"
        dest=/tmp/{{inventory_hostname}}
      when: switch_vendor == 'Juniper Networks'


If I use a simple one level filter such as "filter=groups" or "filter=system" everything works as expected. However if I try to defined a deeper path like "filter=system/syslog" in the configuration I run into problem, getting the following error message:

fatal: [aggrfq178-1.ord6]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to get config: RpcError(severity: error, bad_element: syslog, message: syntax error, expecting </configuration>)"}

In logs I see:

2016-03-05 20:39:46,747:CONFIG:204.93.204.238:connecting to host: root@192.168.1.1:830
2016-03-05 20:39:46,811:ncclient.transport.ssh:Connected (version 2.0, client OpenSSH_6.0)
2016-03-05 20:39:47,259:ncclient.transport.ssh:Authentication (password) successful!
[...]

PR #148, opened 2016-07-27
Update junos_get_config

joobcode:
This is a fix for issue #145. Works for me while current master does not.

This is nice but what I really want is the inverse functionality, e.g exclude parts of the config from archival. Something like the following:

exclude_filter:
              - "interfaces/interface/fxp0"
              - "groups/member0"

This would result in a config being retrieved but minus the sections matching the list of xpath statements. It is also quite confusing that the existing example of "filter" function looks a little like it might be xpath but is not. Well at least it confused me. :-P

### DIFF
--- a/library/junos_get_config
+++ b/library/junos_get_config
@@ -175,8 +175,9 @@ def main():
             current_element = filter_xml
 
             for f in flist:
-                current_element = etree.SubElement(current_element,f)
-
+                current_element.append(etree.SubElement(current_element,f))
+            
+            filter_xml = current_element
             logging.info("Getting config with filter={0}".format(etree.tostring(filter_xml)))
 
         logging.info("Getting config with options={0}".format(options))


joobcode:
Think this is what you wanted @vnitinv

stacywsmith:
@joobcode @vnitinv 

This PR essentially reverts the fix for issue #120 rather than adding the additional functionality mentioned in #145.

While the proposed `exclude_filter` option might be nice, the underlying `<get-configuration>` RPC doesn't support any method of specifying a negative filter.

joobcode:
Ah, I see that now. Oh well, was education for myself if nothing else. :-)

Exclude filter would need to be implemented with the returned xml tree from get_config in pyez rpcmeta. If the <get-configuration> format was text some extra steps would be needed. :-(

stacywsmith:
The new `juniper_junos_config` module now accepts a `filter_xml` option.
The value of the `filter_xml` option can be specified as an XML string which lists multiple configuration hierarchies. (The `filter_xml` option also continues to support a simple XPath-style string which is a directory of a single configuration hierarchy with each component of the hierarchy separated by `/`. i.e. `interfaces/interface/fxp0` gets the `[edit interfaces interface fxp0]` configuration hierarchy.)

Here's an example which retrieves both the `[edit interfaces]` and `[edit protocols]` top-level configuration hierarchies:
```
---
- name: 'Get two sections of Junos config.'
  hosts: junos-all
  connection: local
  gather_facts: no
  roles:
    - Juniper.junos

  tasks:
    - name: 'Get interfaces and protocols config sections'
      juniper_junos_config:
        retrieve: 'candidate'
        filter_xml: '<configuration><interfaces/><protocols/></configuration>'
      register: response
    - name: 'Print result'
      debug:
        var: response
```

Here's selected output from running the playbook:
```
user@h0:~$ ansible-playbook --limit r0 -k pb.test.yaml 
SSH password: 

PLAY [Get two sections of Junos config.] *******************************************************************************

TASK [Get interfaces and protocols config sections] ********************************************************************
ok: [r0]

TASK [Print result] ****************************************************************************************************
ok: [r0] => {
    "response": {
        "changed": false, 
        "config": "\n## Last changed: 2017-11-10 14:30:58 PST\ninterfaces {\n    ge-0/0/0 {\n        unit 
... output trimmed ...
        "config_lines": [
            "", 
            "## Last changed: 2017-11-10 14:30:58 PST", 
            "interfaces {", 
            "    ge-0/0/0 {", 
            "        unit 0 {", 
            "            family inet {", 
            "                address 10.0.4.1/30;", 
            "            }", 
            "            family iso;", 
            "        }", 
            "    }", 
            "    ge-0/0/1 {", 
            "        unit 0 {", 
            "            family inet {", 
            "                address 10.0.2.1/30;", 
            "            }", 
            "            family iso;", 
            "        }", 
            "    }", 
            "    ge-0/0/2 {", 
            "        unit 0 {", 
            "            family inet {", 
            "                address 10.0.3.1/30;", 
            "            }", 
            "            family iso;", 
            "        }", 
            "    }", 
            "    ge-0/0/3 {", 
            "        unit 0 {", 
            "            family inet {", 
            "                address 10.0.5.1/30;", 
            "            }", 
            "            family iso;", 
            "        }", 
            "    }", 
            "    ge-0/0/4 {", 
            "        unit 0 {", 
            "            family inet {", 
            "                address 10.0.1.1/30;", 
            "            }", 
            "            family iso;", 
            "        }", 
            "    }", 
            "    lo0 {", 
            "        unit 0 {", 
            "            family inet {", 
            "                address 10.210.255.200/32 {", 
            "                    primary;", 
            "                }", 
            "            }", 
            "            family iso {", 
            "                address 47.0005.80ff.f800.0000.0108.0001.0102.1025.5200.00;", 
            "            }", 
            "            family inet6 {", 
            "                address abcd::10:210:255:200/128 {", 
            "                    primary;", 
            "                }", 
            "            }", 
            "        }", 
            "    }", 
            "}", 
            "## Last changed: 2017-11-10 14:30:58 PST", 
            "protocols {", 
            "    isis {", 
            "        apply-groups ISIS-INTERFACE;", 
            "        traceoptions {", 
            "            file isis.log size 200m files 3;", 
            "            flag error;", 
            "            flag state;", 
            "        }", 
            "        level 1 disable;", 
            "        level 2 wide-metrics-only;", 
            "        interface ge-0/0/0.0 {", 
            "            level 2 metric 10;", 
            "        }", 
            "        interface ge-0/0/1.0 {", 
            "            level 2 metric 10;", 
            "        }", 
            "        interface ge-0/0/2.0 {", 
            "            level 2 metric 10;", 
            "        }", 
            "        interface ge-0/0/3.0 {", 
            "            level 2 metric 10;", 
            "        }", 
            "        interface ge-0/0/4.0 {", 
            "            level 2 metric 10;", 
            "        }", 
            "        interface lo0.0 {", 
            "            level 2 {", 
            "                metric 0;", 
            "                passive;", 
            "            }", 
            "        }", 
            "    }", 
            "    lldp {", 
            "        port-id-subtype interface-name;", 
            "        interface all;", 
            "        interface fxp0 {", 
            "            disable;", 
            "        }", 
            "    }", 
            "}"
        ], 
        "failed": false, 
        "msg": "Configuration has been: opened, retrieved, closed."
    }
}

PLAY RECAP *************************************************************************************************************
r0                         : ok=2    changed=0    unreachable=0    failed=0   
```

PR #151, opened 2016-07-31
junos_jsnapy : add tests to check if test files are present 

dgarros:


dgarros:
@vnitinv / @stacywsmith  please can you check if it's good for you
I wanted to add similar check for config_file but I couldn't test easily, should I add it ?

vnitinv:
@dgarros I am fine with the changes.

PR #153, opened 2016-08-08
fix regex to match xe/ge/et interfaces

stianvi:
I need the regex to match all the following:
```
kwargs=vme
kwargs=xe-0/0/0
kwargs=fpx0
kwargs=me0.0
```
Current state don't match xe/ge/et itnerfaces due to forward slash not in the regex expression.


stianvi:
See Juniper/py-junos-eznc#549

stianvi:
**before**
```
$ cat get_interface_information-et-var.conf

protocol: operation-failed
error: device et-0 not found
```

**after**
```
$ cat get_interface_information-et.conf

Physical interface: et-0/1/0, Enabled, Physical link is Up
  Interface index: 183, SNMP ifIndex: 512
```

PR #165, opened 2016-09-20
Add Dockerfile

dgarros:
I created a new Dockerfile inspired by @ntwrkguru's work to automatically build a docker container for the ansible modules, it's very similar to what we have done with Pyez

@vnitinv, please review


ntwrkguru:
Now that it's been merged, we need to update the repo on the Hub. @dgarros, I couldn't figure out how to change the repo. Any clue?

dgarros:
Done, 
BTW, should we change the name of the container ?
Not sure if `juniper/pyez-ansible` is the most appropriate name
I would suggest something like `juniper/ansible-galaxy`

@ntwrkguru, @vnitinv  Thoughts ?

ntwrkguru:
How about just `juniper/ansible-junos`?

PR #167, opened 2016-09-23
junos_rpc should always return rpc_reply

spidercensus:
This is feature is already provided by the junos_command core module. We should be able to do the same in junos_rpc.


vnitinv:
@spidercensus 

As below code has been removed

-                else:
 -                    if args.get('format') == 'json':
 -                        results['rpc_reply'] = rpc_reply
 -                        confile.write(str(rpc_reply))

I feel this pull request will break copying data in output file.

stacywsmith:
@spidercensus This functionality have been implemented in the new `juniper_junos_rpc` module.

The RPC reply is now returned as part of the response. It may also still be written to a file if the `dest` or `dest_dir` option is specified.

See [these lines](https://github.com/Juniper/ansible-junos-stdlib/blob/master/library/juniper_junos_rpc.py#L314-L334) for details on the return values from the module.

PR #168, opened 2016-09-27
Setting commit_onwarnings to True allows commiting even if warnings "statement not found" are raised

ymitsos:
In certain cases one may prefer to create a configuration snippet that firstly deletes a part of the respective system configuration and then applies the desired functionality. In such cases warning statements of type "statement not found" may occur because prevailing delete commands may not correspond to existing configuration if the configuration templates are generic enough to cope with a variety of roles/configurations.

It is therefore, useful to proceed with loading the configuration to the devices even if such warnings are raised. For any other type of warning statement or error message, the functionality remains untouched; namely, fails with the appropriate message.

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -150,6 +150,14 @@ options:
         required: false
         default: None
         choices: [1 - 4]
+    commit_onwarnings:
+        description:
+            - Ignore "statement not found" warning messages during loading 
+              configuration. Proceed with commit. 
+        required: false
+        default: no
+        choices: ['true','false','yes','no']
+
 '''
 
 EXAMPLES = '''
@@ -220,6 +228,7 @@ def _load_via_netconf(module):
     in_check_mode = module.check_mode
     overwrite = module.boolean(module.params['overwrite'])
     replace = module.boolean(module.params['replace'])
+    commit_onwarnings = module.boolean(module.params['commit_onwarnings'])
 
     if all([overwrite, replace]):
         msg = "Overwrite and Replace cannot both be True!"
@@ -275,7 +284,27 @@ def _load_via_netconf(module):
         elif overwrite is False:
             load_args['merge'] = True
         cu.load(**load_args)
-    except (ValueError, ConfigLoadError, Exception) as err:
+    # If only "statement not found" warning messages are raised and 
+    # commit_onwarnings is set to "True", commit proceeds, otherwise fails. 
+    except ConfigLoadError as err:
+        severity = list(error['severity'] for error in err.errs)
+        messages = list(error['message'] for error in err.errs)
+
+        if not ((len(set(severity)) <=1 and severity[0] == 'warning') and
+                (len(set(messages)) <=1 and messages[0] == 'statement not found') and
+                commit_onwarnings):
+            msg = "Unable to load config: {0}".format(err)
+            logging.error(msg)
+            try:
+                logging.info("unlocking")
+                cu.unlock()
+            except UnlockError as err:
+                logging.error("Unable to unlock config: {0}".format(err))
+            dev.close()
+            module.fail_json(msg=msg)
+        else:
+            logging.info("Warnings 'statement not found' were raised but ignored, proceeding with commit.")
+    except (ValueError, Exception) as err:
         msg = "Unable to load config: {0}".format(err)
         logging.error(msg)
         try:
@@ -433,7 +462,8 @@ def main():
             port=dict(required=False, default=830),
             mode=dict(required=False, default=None),
             confirm=dict(required=False, default=None),
-            check_commit_wait=dict(required=False, default=None)
+            check_commit_wait=dict(required=False, default=None),
+            commit_onwarnings=dict(required=False, type='bool', choices=BOOLEANS, default=False)
         ),
         supports_check_mode=True)
 


pamelia:
Very useful when dealing with a set/delete statement based config.

ymitsos:
any thoughts on this?

ydnath:
@stacywsmith to comment

ymitsos:
@stacywsmith did you have time to review it?

vnitinv:
@ymitsos @stacywsmith Stacy, this looks good to me. Please do review.

library/junos_install_config
```diff
@@ -150,6 +150,14 @@ options:
         required: false
         default: None
         choices: [1 - 4]
+    commit_onwarnings:
```
stacywsmith: I think the name of the argument is a misnomer. It doesn't commit on all warnings, just "statement not found" warnings. Therefore, I suggest we change the name of the argument to `ignore_statement_not_found`.

library/junos_install_config
```diff
@@ -220,6 +228,7 @@ def _load_via_netconf(module):
     in_check_mode = module.check_mode
     overwrite = module.boolean(module.params['overwrite'])
     replace = module.boolean(module.params['replace'])
+    commit_onwarnings = module.boolean(module.params['commit_onwarnings'])
```
stacywsmith: Change `commit_onwarnings` to `ignore_statement_not_found`.

library/junos_install_config
```diff
@@ -275,7 +284,27 @@ def _load_via_netconf(module):
...
-    except (ValueError, ConfigLoadError, Exception) as err:
+    # If only "statement not found" warning messages are raised and 
+    # commit_onwarnings is set to "True", commit proceeds, otherwise fails. 
+    except ConfigLoadError as err:
+        severity = list(error['severity'] for error in err.errs)
+        messages = list(error['message'] for error in err.errs)
+
+        if not ((len(set(severity)) <=1 and severity[0] == 'warning') and
+                (len(set(messages)) <=1 and messages[0] == 'statement not found') and
+                commit_onwarnings):
```
stacywsmith: Change `commit_onwarnings` to `ignore_statement_not_found`.

library/junos_install_config
```diff
@@ -275,7 +284,27 @@ def _load_via_netconf(module):
...
+            logging.error(msg)
+            try:
+                logging.info("unlocking")
+                cu.unlock()
+            except UnlockError as err:
+                logging.error("Unable to unlock config: {0}".format(err))
+            dev.close()
+            module.fail_json(msg=msg)
+        else:
+            logging.info("Warnings 'statement not found' were raised but ignored, proceeding with commit.")
```
stacywsmith: Change from `"Warnings 'statement not found' were raised but ignored,` to `"'statement not found' warnings were raised but ignored,`

library/junos_install_config
```diff
@@ -433,7 +462,8 @@ def main():
             port=dict(required=False, default=830),
             mode=dict(required=False, default=None),
             confirm=dict(required=False, default=None),
-            check_commit_wait=dict(required=False, default=None)
+            check_commit_wait=dict(required=False, default=None),
+            commit_onwarnings=dict(required=False, type='bool', choices=BOOLEANS, default=False)
```
stacywsmith: Change `commit_onwarnings` to `ignore_statement_not_found`.

library/junos_install_config
```diff
@@ -275,7 +284,27 @@ def _load_via_netconf(module):
...
             load_args['merge'] = True
         cu.load(**load_args)
-    except (ValueError, ConfigLoadError, Exception) as err:
+    # If only "statement not found" warning messages are raised and 
+    # commit_onwarnings is set to "True", commit proceeds, otherwise fails. 
+    except ConfigLoadError as err:
+        severity = list(error['severity'] for error in err.errs)
+        messages = list(error['message'] for error in err.errs)
+
+        if not ((len(set(severity)) <=1 and severity[0] == 'warning') and
```
stacywsmith: This conditional block seems incorrect. If there are multiple 'statement not found' warnings, this conditional would be true and cause the commit to fail.

ymitsos: @stacywsmith the conditional block is correct and I validated once again. Let me know if I need to elaborate on this.

jeffbrl:
@ymitsos can you please open an issue to document the issue that this pull request addresses? It's easier for people who encounter problems to find open issues rather than PRs in progress. I almost opened an issue last night when I observed this behavior and just happened to find this PR before I hit submit.

ymitsos:
@jeffbrl you are right, just filled an issue https://github.com/Juniper/ansible-junos-stdlib/issues/201

XioNoX:
Would it be possible to include the warning: "statement has no contents; ignored" as well?
I have a similar workflow, but instead of using set commands I use the bracket form.
For example:
```
protocols {
    replace: bgp {
      {% if bgp.peers | d([]) %}
      {% endif %}
    }
}
```
This allows me to  remove all the BGP configuration if "bgp.peers" is not set.
Inserting replace: bgp {} inside the if statement would lead to stale config if bgp.peers isn't defined.

Or more globally, it might make more sens to be able to set a list of warning messages to be ignore?

bhanu04:
Hi,
i am looking the same feature in my playbook. may i know if this feature is available to use?

Thanks
Bhanu

stacywsmith:
This has now been implemented via the `ignore_warning` option on the new `juniper_junos_config` module via #290.

Documentation for the `ignore_warning` option is:
```
  ignore_warning:
    description:
      - A boolean, string or list of strings. If the value is C(true),
        ignore all warnings regardless of the warning message. If the value
        is a string, it will ignore warning(s) if the message of each warning
        matches the string. If the value is a list of strings, ignore
        warning(s) if the message of each warning matches at least one of the
        strings in the list. The C(ignore_warning) value is applied to the
        load and commit operations performed by this module.
        load and commit operations performed by this module.
    required: false
    default: none
    type: bool, str, or list of str
```

PR #176, opened 2016-11-15
Fix missing interpreter: (on Rhel7)

yurilaaziz:
We get this error last week on our Rhel7 server.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 from setuptools import setup
 from version import VERSION
 


stacywsmith:
@yurilaaziz 

This fix makes sense for the setup.py file in order for the modules to be correctly installed. I will merge this pull request. Thanks for submitting it.

This isn't directly applicable to your commit, but it's worth stating here for future reference.

Using `/usr/bin/env` is not the recommended approach within modules themselves. For modules, the shebang line of the modules should be set to `#!/usr/bin/python` and the `ansible_python_interpreter` variable should be used as needed to specify an non-default Python2 location.

See: [this reference](http://docs.ansible.com/ansible/faq.html#how-do-i-handle-python-pathing-not-having-a-python-2-x-in-usr-bin-python-on-a-remote-machine) or
[that reference](http://docs.ansible.com/ansible/dev_guide/developing_modules.html#module-checklist)

The modules don't currently comply with this recommendation. They currently have `/usr/bin/env` in the shebang line instead. We should probably address that under a separate issue.

Issue #190, opened 2016-12-07
coercing to Unicode: need string or buffer, NoneType found


blc325:
Getting this fatal response when trying to execute playbook below on an EX4300 virtual-chassis:
TASK [perform a system snapshot] ***********************************************
fatal: [10.199.244.247]: FAILED! => {"changed": false, "failed": true, "msg": "Uncaught exception - please report: coercing to Unicode: need string or buffer, NoneType found"}

      - name: perform a system snapshot
        junos_cli:
          host: "{{ inventory_hostname }}"
          user: "{{username}}"
          passwd: "{{userpasswd}}"
          cli: "request system snapshot slice alternate all-members"
          timeout: 600
        register: system_snapshot

dgarros:
Hi 
Please can you provide more information to help with the troubleshooting
- Output of `request system snapshot slice alternate all-members | display xml`
- Ansible cli logs with high verbosity `-vvv`

thanks
Damien

blc325:
I think it could be because there are long pauses between different parts of the message before EOF.



On Dec 6, 2016, at 10:53 PM, Damien <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi
Please can you provide more information to help with the troubleshooting

  *   Output of request system snapshot slice alternate all-members | display xml
  *   Ansible cli logs with high verbosity -vvv

thanks
Damien

blc325:
Here’s an example of the XML response with long pauses between FPCs.


Last login: Mon Nov 28 14:17:36 2016 from 192.168.254.1

brianchow@ubuntu:~$ ssh root@vc

Password:

--- JUNOS 14.1X53-D35.3 built 2016-03-01 02:32:19 UTC

root@4300VC:RE:1% cli



{master:1}

root@4300VC> request system snapshot slice alternate all-members | display xml

<rpc-reply xmlns:junos="http://xml.juniper.net/junos/14.1X53/junos">

    <multi-routing-engine-results>



        <multi-routing-engine-item>



            <re-name>fpc0</re-name>



            <snapshot-information>

                <inform>

                    <message>Formatting alternate root (/dev/da0s2a)...</message>

                </inform>

                /dev/da0s2a: 329.3MB (674472 sectors) block size 16384, fragment size 2048

                        using 4 cylinder groups of 82.34MB, 5270 blks, 10624 inodes.

                super-block backups (for fsck -b #) at:

                 32, 168672, 337312, 505952<snapshot-copy>

[...]

PR #191, opened 2016-12-07
Require the dest argument to the junos_cli module.

stacywsmith:
Addresses #190.

### DIFF
--- a/library/junos_cli
+++ b/library/junos_cli
@@ -133,7 +133,7 @@ def main():
                            mode=dict(required=False, default=None),
                            timeout=dict(required=False, type='int', default=0),
                            logfile=dict(required=False, default=None),
-                           dest=dict(required=False, default=None),
+                           dest=dict(required=True, default=None),
                            format=dict(required=False, choices=['text', 'xml'], default='text')
                            ),
         supports_check_mode=True)


stacywsmith:
@dgarros Can you please take a look at this?

dgarros:
Hi @stacywsmith 
There are some commands that return nothing hence I don't think we should make `dest` mandatory.
have you found an issue because dest was missing ?

stacywsmith:
@dgarros Yes, if dest is not specified an AttributeError exception is raised. Please see the discussion on #190.

dgarros:
got it, in this case we should handle the exception properly or check if `dest` is defined.   
I'll give it a try quickly

stacywsmith:
@dgarros The issue is [here](https://github.com/Juniper/ansible-junos-stdlib/blob/master/library/junos_cli#L184). We attempt to open the file specified by `dest` even if the value is `None`. I think you can simply wrap this block in an if:
```
        if dest is not None:
            with open(args['dest'], 'w') as outputfile:
                if args['format'] == 'text':
                    outputfile.write(cli_output)
                elif args['format'] == 'xml':
                    outputfile.write(etree.tostring(cli_output))
```

We also need to clean up the conflicting documentation which says the `dest` argument is required.

stacywsmith:
Handled by #193.

PR #192, opened 2016-12-09
junos_rpc: rpc_reply string is null for text or XML formats

dorkmatt:
The `junos_rpc` module appears to only populate the `.rpc_reply` results string for JSON output. For those of not running JunOS >14, it's still useful to get this out - even if text or XML format. The goal here is to use this data in another play, no need to output to a file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN tar -czf Juniper.junos ansible-junos-stdlib &&\
     pip install junos-netconify &&\
     pip install jxmlease &&\
     pip install ansible &&\
+    pip install jsnapy &&\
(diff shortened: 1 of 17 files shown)


dorkmatt:
I thought my indenting was catching all cases, apologies - not sure how to fix this up to catch all scenarios then?

stacywsmith:
This functionality have been implemented in the new `juniper_junos_rpc` module.

The RPC reply is now returned as part of the response. It may also still be written to a file if the `dest` or `dest_dir` option is specified.

See [these lines](https://github.com/Juniper/ansible-junos-stdlib/blob/master/library/juniper_junos_rpc.py#L314-L334) for details on the return values from the module.

PR #193, opened 2016-12-13
junos_cli / Add test before writing output to dest

dgarros:
fix issue with #191


dgarros:
@stacywsmith please can you check

PR #206, opened 2017-01-28
Add jsnapy folder argument

XioNoX:
Add a way to indicate jsnapy where to find all the testfiles, snapshots, and jsnapy main config file.

jsnapy.cfg requires absolute paths in the [DEFAULT] section, which is not possible in most cases as users will have their playbook in different folders.
Being able to specify the jsnapy "folder" variable, allows to use ansible variables (and thus relative to the playbook).
FYI, the variable is being used there:
https://github.com/Juniper/jsnapy/blob/0891a3b0d88f67fe7ff028dd0b77ecbb96918310/lib/jnpr/jsnapy/__init__.py#L31

### DIFF
--- a/library/junos_jsnapy
+++ b/library/junos_jsnapy
@@ -99,6 +99,11 @@ options:
             - Path for the JSNAPy yaml testfiles/configuration file
         required: false
         default: '/etc/jsnapy/testfiles'
+    folder:
+        description:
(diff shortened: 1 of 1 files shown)


vnitinv:
@XioNoX This feature was added when we revamped our modules in 2.0.x. Hence closing this pull request. Thanks for your contribution.

PR #212, opened 2017-03-08
Updated Dockerfile and README

ntwrkguru:


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM juniper/pyez:2.0.1
-MAINTAINER ntwrkguru@gmail.com
+FROM juniper/pyez:latest
+MAINTAINER Stephen Steiner <ssteiner@juniper.net>
 
-ARG ver_ansible=2.2.0.0
+ARG ver_ansible=2.3.0.0
```
stacywsmith: At the moment, these modules are not working with the development version of Ansible 2.3 due to some bugs in Ansible. Can we please hold off on using Ansible 2.3?

PR #217, opened 2017-03-16
Include openssh-client library

jorgebonilla:
In order to be able to install keys to connect using "ssh-add ˜/.ssh/key"
Added   apk add openssh-client &&\


stacywsmith:
@jorgebonilla Can you please take a look at and address the failure with Ansible 2.1.3?
https://travis-ci.org/Juniper/ansible-junos-stdlib/jobs/211579156

Issue #221, opened 2017-04-05
junos_get_table with response_type=juniper_items ommits information


ymitsos:
response_type flag defines how data are returned upon execution of the junos_get_table module. User may select either 'list_of_dicts', which is the default value, or 'juniper_items'. However, the latter returns the "key" value used in the Table definition, while the former not.

For instance, if we'd like to fetch ISIS statistics, we may define the table as follows:
```
IsisStatisticsTable:
  rpc: get-isis-statistics-information
  item: //isis-pdu-statistics
  key:
    - ./isis-pdu-type
  view: IsisStatisticsView
```
setting response_type=list_of_dicts will return, among others: 
```
...
"resource": [
                                {
                                    "packets_dropped": "12345", 
                                    "packets_processed": "12345", 
                                    "packets_received": "12345", 
                                    "packets_retransmitted": "12345", 
                                    "packets_sent": "12345"
                                }, 
                                {
...
```
while setting response_type=juniper_items will return: 
```
...
 "resource": [
                                [
                                    [
                                        "LSP"
                                    ], 
                                    [
                                        [
                                            "packets_processed", 
                                            "12345"
                                        ], 
                                        [
                                            "packets_sent", 
                                            "12345"
                                        ], 
                                        [
                                            "packets_retransmitted", 
                                            "12345"
                                        ], 
                                        [
                                            "packets_received", 
                                            "12345"
                                        ], 
                                        [
                                            "packets_dropped", 
                                            "12345"
                                        ]
                                    ]
                                ], 

...
```
The first item of the list ("LSP" in this case) is the key with which we performed the parsing of the table data. This information is not available in the default behavior of this module. Given that this information is available in the RPC call return value, it would be nice if it could be embedded in the list of dicts returned to ansible and to the user.

stacywsmith:
Please set my response to #225.

PR #225, opened 2017-04-08
junos_get_table includes the table's key in response

ymitsos:
junos_get_table with response_type set to list_of_dicts returns the key of the table within the returned list of dictionaries as well, which is the default behavior of the "juniper_items" mode.

Relates to issue: https://github.com/Juniper/ansible-junos-stdlib/issues/221

### DIFF
--- a/library/junos_get_table
+++ b/library/junos_get_table
@@ -177,7 +177,7 @@ def juniper_items_to_list_of_dicts(data):
             if normalized_value and not isinstance(normalized_value, (str, list, int, dict)):
                 normalized_value = juniper_items_to_list_of_dicts(normalized_value)
             temp[normalized_key] = normalized_value
-        list_of_resources.append(temp)
+        list_of_resources.append({table_key[0]: temp})
     return list_of_resources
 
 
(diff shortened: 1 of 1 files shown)


stacywsmith:
@ymitsos Unfortunately, your proposed change modifies the returned data structure in a way that is not backwards compatible. I actually prefer your (list_of_dict_of_dicts) format which includes the key value. However, I think it would be unwise to break existing playbooks in order to add this functionality.

In many cases your proposed change is also not strictly necessary. That's because many Table/View definitions include the item key in the view, as in the `LLDPNeighborTable` example below:
```
---
LLDPNeighborTable:
  rpc: get-lldp-neighbors-information
  item: lldp-neighbor-information
  key: lldp-local-interface | lldp-local-port-id
  view: LLDPNeighborView

LLDPNeighborView:
  fields:
    local_int: lldp-local-interface | lldp-local-port-id
    local_parent: lldp-local-parent-interface-name
    remote_type: lldp-remote-chassis-id-subtype
    remote_chassis_id: lldp-remote-chassis-id
    remote_port_desc: lldp-remote-port-description
    remote_port_id: lldp-remote-port-id
    remote_sysname: lldp-remote-system-name
```

A backwards compatible option would be to support a third `list_of_dict_of_dicts` value for the `response_type` option in the reworked `juniper_junos_table` module. If you want to submit a pull request with that functionality, I will review and merge it.

Alternatively, I suggest simply adding the item key as a field in the view when defining any custom Table/Views.

Issue #233, opened 2017-04-25
Junos_Install_Os Failed to upgrade the JUNOS Host Software


nzkller:
Running junos_install_os will successfully install the Junos Base OS Software Suite but its does not upgrade the Juniper Host Software in Juniper's QFX Series switches.

Juniper's Advanced JTAC Agent specified that this might create issues in the performance of the switch and it could be responsible for some of the problem we were having.

Example:

Hostname: tesw1
Model: qfx5100-48t-6q
Junos: 14.1X53-D30.3
JUNOS Base OS Software Suite [14.1X53-D30.3]
JUNOS Base OS boot [14.1X53-D30.3]
JUNOS Crypto Software Suite [14.1X53-D30.3]
JUNOS Online Documentation [14.1X53-D30.3]
JUNOS Kernel Software Suite [14.1X53-D30.3]
JUNOS Packet Forwarding Engine Support (qfx-ex-x86-32) [14.1X53-D30.3]
JUNOS Routing Software Suite [14.1X53-D30.3]
JUNOS SDN Software Suite [14.1X53-D30.3]
JUNOS Enterprise Software Suite [14.1X53-D30.3]
JUNOS Web Management Platform Package [14.1X53-D30.3]
JUNOS py-base-i386 [14.1X53-D30.3]
**JUNOS Host Software [13.2X51-D25.2]** 

Normally, if I'd perform the upgrade by hand I will use the command

`request system software add reboot no-validate force-host /var/tmp/jinstall.......`

But I need this to be perform by the Ansible playbook that will provision new racks

Is there anyway to make the junos_install_os upgrade the host software as well?

Thanks in advance and best regards,

Nzkller

stacywsmith:
Hi @nzkller 

Sorry, I'm not familiar enough with the specific nuances of the QFX platform, so I have a couple of questions for you.

When upgrading these devices manually from the CLI, is there any difference in the actual package file that you supply to the `request system software add` command? If not, is the only difference the addition of the `force-host` argument to the command?

nzkller:
Hi @stacywsmith,

Yes you are absolutely right the only difference is the force-host argument that you add at the end. 

Thank you for your reply and best regards,

Nzkller

stacywsmith:
Thanks @nzkller 

It's not currently possible to pass the `force-host` argument via the `junos_install_os` module, but I will work on adding that capability.

Would you mind emailing me your JTAC case number to stacys at juniper dot net?

[...]

PR #234, opened 2017-04-25
Add the force_host argument to junos_install_os. Addresses #233.

stacywsmith:
By default, the Host Software package on QFX-series platforms is not
upgraded by the `<request-package-add>` RPC. Adding the `<force-host/>`
argument to the RPC forces the Host Software package to also be upgraded.

This PR adds the ability to pass the `force_host` argument to the
`junos_install_os` module. This optional boolean argument defaults
to `False`. When set to `True`, it passes the `<force-host/>` argument to
the underlying `<request-package-add>` RPC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,12 @@ docs/*.rst
 docs/_build/
 
 tests/.vagrant/*
+
+# Window File Explorer
+desktop.ini
+
+# Mac OS X Finder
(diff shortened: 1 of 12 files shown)


stacywsmith:
@vnitinv Just pinging to see if you can please review this change. It's fairly simple. Thanks.

PR #238, opened 2017-05-04
Add the options argument to the junos_install_os module

stacywsmith:
These opaque options get passed through to PyEZ and eventually
the `<request-package-add>` RPC.

### DIFF
--- a/library/junos_install_os
+++ b/library/junos_install_os
@@ -108,6 +108,13 @@ options:
         required: false
         default: false
         choices: ['true','false', 'yes','no']
+    force_host:
+        description:
+             - If set to B(true), forces the upgrade of the Host Software 
+               package on QFX-series devices.
+        required: false
+        default: false
+        choices: ['true','false', 'yes','no']
     reboot_pause:
         description:
             - Amount of time in seconds to wait after the reboot is issued
@@ -192,7 +199,7 @@ def junos_install_os(module, dev):
         def do_log(msg, level='info'):
             getattr(logging, level)(msg)
     else:
-        def do_log(msg):
+        def do_log(msg, level=None):
             pass
 
     # -------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


library/junos_install_os
```diff
@@ -265,23 +294,17 @@ def main():
...
+        if not LooseVersion(VERSION) >= LooseVersion('2.1.2'):
+            module.fail_json(msg='junos-eznc >= 2.1.2 is required for this '
+                                 'module')
     except ImportError as ex:
         module.fail_json(msg='ImportError: %s' % ex.message)
 
-    if int(args['port']) == 23 and LooseVersion(VERSION) < LooseVersion('2.0.0'):
-        module.fail_json(msg='junos-eznc >= 2.0.0 is required for telnet connection.')
-
-    if (module.boolean(args['issu']) or module.boolean(args['nssu'])) and \
```
vnitinv: Have you done the changes on top of latest code. Why issu/nssu check has been removed?

library/junos_install_os
```diff
@@ -155,7 +161,20 @@ import logging
 import re
 import os
 from distutils.version import LooseVersion
+from ansible.module_utils.six import string_types
+from ansible.module_utils.basic import *
 
+def _convert_to_boolean(val):
+    if val is None or isinstance(val, bool):
+        return val
```
vnitinv: I think we can keep this function functionality inside the for loop itself for 2 reasons:
1> we are returning None in first if condition and None is not a boolean
2> Raising a typeError just to capture other data can be removed if we have code in for loop itself

library/junos_install_os
```diff
@@ -155,7 +161,20 @@ import logging
...
+    if isinstance(val, string_types):
+        val = val.lower()
+    if val in BOOLEANS_TRUE:
+        return True
+    elif val in BOOLEANS_FALSE:
+        return False
+    else:
+        raise TypeError
```
vnitinv: I dont its good idea to raise error for other data type and then catch error to set non boolean value in line 247.

stacywsmith:
The new `juniper_junos_software` module introduced in #296 now exposes all of the PyEZ `install()` method keywords via Ansible options. In addition, it supports a `kwargs` option for passing arbitrary keywords and values through to the underlying RPC.

PR #248, opened 2017-06-19
ignore_warning support for junos_install_config

vnitinv:
user can now pass ignore_warning as per https://github.com/Juniper/py-junos-eznc/blob/master/lib/jnpr/junos/utils/config.py#L336

```yaml
  tasks:
    - name: install config to device
      junos_install_config:
        ignore_warning: True
        file: xxxx.set
```
```yaml
  tasks:
    - name: install config to device
      junos_install_config:
        ignore_warning: "statement not found"
        file: xxxx.set
```
```yaml
  tasks:
    - name: install config to device
      junos_install_config:
        ignore_warning:
          - "statement not found"
          - "testing"
        file: xxxx.set
```

### DIFF
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -261,7 +261,7 @@ def _load_via_netconf(module):
         logging.error(msg)
         module.fail_json(msg=msg)
 
-    if args['ignore_warning'] in BOOLEANS + ["True", "False"]:
+    if str(args['ignore_warning']).lower() in BOOLEANS:
         args['ignore_warning'] = module.boolean(module.params['ignore_warning'])
     elif isinstance(args['ignore_warning'], str) and \
             re.search('\[.*\]', args['ignore_warning']):
--- a/library/junos_install_config
+++ b/library/junos_install_config
@@ -51,7 +51,7 @@ description:
       NETCONF include ASCII text, Junos XML elements, and Junos OS B(set) commands.
       Configurations performed through the console must only use ASCII text formatting.
 requirements:
-    - junos-eznc >= 1.2.2
+    - junos-eznc >= 2.1.1
     - junos-netconify >= 1.0.1, when using the I(console) option
 options:
     host:
@@ -223,7 +223,10 @@ import logging
 from os.path import isfile
 import os
 from distutils.version import LooseVersion
+import re
 import ast
+
+from ansible.module_utils.basic import *
 
 
 def _load_via_netconf(module):
@@ -235,13 +238,10 @@ def _load_via_netconf(module):
             ConfigLoadError, CommitError
         from jnpr.junos.utils.config import Config
         from jnpr.junos.version import VERSION
-        if not LooseVersion(VERSION) >= LooseVersion('1.2.2'):
-            module.fail_json(msg='junos-eznc >= 1.2.2 is required for this module')
+        if not LooseVersion(VERSION) >= LooseVersion('2.1.1'):
+            module.fail_json(msg='junos-eznc >= 2.1.1 is required for this module')
     except ImportError as ex:
         module.fail_json(msg='ImportError: %s' % ex.message)
-
-    if args['mode'] is not None and LooseVersion(VERSION) < LooseVersion('2.0.0'):
-        module.fail_json(msg='junos-eznc >= 2.0.0 is required for console connection')
 
     logfile = args['logfile']
     if logfile is not None:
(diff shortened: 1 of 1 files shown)


library/junos_install_config
```diff
@@ -250,6 +261,9 @@ def _load_via_netconf(module):
         logging.error(msg)
         module.fail_json(msg=msg)
 
+    if args['ignore_warning'] in BOOLEANS + ["True", "False"]:
```
stacywsmith: Ansible boolean values are normally case-insensitive, but this is not. Suggest changing to:
```
if str(args['ignore_warning']).lower() in BOOLEANS + ["true", "false"]:
```

library/junos_install_config
```diff
@@ -250,6 +261,9 @@ def _load_via_netconf(module):
...
+        args['ignore_warning'] = module.boolean(module.params['ignore_warning'])
+
```
stacywsmith: It would be helpful to do additional arg type checking here to ensure that `args['ignore_warning']` is a string type of list of string types.

library/junos_install_config
```diff
@@ -165,6 +165,16 @@ options:
...
         required: false
         default: None
+    ignore_warning:
+        description:
+            - A boolean, string or list of string.
+          If the value is True, it will ignore all warnings regardless of the
+          warning message. If the value is a string, it will ignore
+          warning(s) if the message of each warning matches the string. If
+          the value is a list of strings, ignore warning(s) if the message of
+          each warning matches at least one of the strings in the list.
```
stacywsmith: Add a line stating:
```
All warning messages produced when loading, diffing, checking, or committing the configuration are checked against ignore_warning. 
```

Then add code to pass ignore_warning value when diffing or checking the config.

vnitinv: I dont think we should do it for `diff` as I never diff function is PyEZ itself doesnt support that and also I never encountered diff giving rpc-error (with severity Warning). I dont see any reason too for getting such error from diff.
commit_check we can do, but for this function PyEZ don't support for now. So we cant do changes as of today.

library/junos_install_config
vnitinv: I have not defined ignore_warning as a string for few reasons. With the older version of Ansible type will be taken care of. So if I pass in Boolean, string, list it will remain so.
Do you suggest me to make ignore_warning type defined as string?

library/junos_install_config
vnitinv: I didnt do that for few reason:
older version of ansible will have it as list itself. with >=2.3 it will be string and in PyEZ we do regular expression check where we loop with all rpc-warning and check if that matches any on in the list.
In this case it will just search/match the warnings in the string of list (for ex "['statment not found', 'xyz']"

library/junos_install_config
vnitinv: @stacywsmith I have tested all possible scenarios and dont see any failure case. Can you help me identifying the condition in which we would require this?

library/junos_install_config
vnitinv: @stacywsmith converting string (of list) to list now.

artcrime:
Hi,
when will this change be commited? We have issues pushing a configuration to a SRX300 with Junos 15.1 cause a master password is set an the passwords in our .set configuration are not complex enough. So the ansible module will not commit the config because of the warnings.

vnitinv:
@stacywsmith can you please review latest changes to this pull request. We need to check why Ravelo is now failing with ConnectUnknownHostError

library/junos_install_config
stacywsmith: Actually, we have encountered diff returning a warning.
https://github.com/Juniper/py-junos-eznc/blob/master/lib/jnpr/junos/utils/config.py#L228-L233

We did that before we added `ignore_warning` to PyEZ.

vnitinv: @stacywsmith Even if we want to do ignore_warning support for diff function then we first need to change the PyEZ diff function itself. Right now it won't accept any extra parameters.

https://github.com/Juniper/py-junos-eznc/blob/master/lib/jnpr/junos/utils/config.py#L208

stacywsmith: @vnitinv Yes. Agreed. If we need ignore_warning support for diff, we can add it at a future point.

library/junos_install_config
```diff
@@ -250,11 +261,21 @@ def _load_via_netconf(module):
         logging.error(msg)
         module.fail_json(msg=msg)
 
-    logging.info("connecting to host: {0}@{1}:{2}".format(args['user'], args['host'], args['port']))
+    if args['ignore_warning'] in BOOLEANS + ["True", "False"]:
```
stacywsmith: The idea here is to match the exact same things that Ansible would normally interpret as a boolean.

From http://docs.ansible.com/ansible/YAMLSyntax.html:
>Ansible doesn’t really use these too much, but you can also specify a boolean value (true/false) in several forms:

create_key: yes
needs_agent: no
knows_oop: True
likes_emacs: TRUE
uses_cvs: false

The current code would not match the string `TRUE`, for example.
The Ansible code for determining boolean values does a case-insensitive match on the boolean values. You can see this here:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L2046-L2057

That's why I'm suggesting this line should be:
```
if str(args['ignore_warning']).lower() in BOOLEANS + "none":
```
That should match the same values which `module.boolean()` will actually convert into a boolean value.

vnitinv: Stacy, Why is + "none" is required here?

stacywsmith: I was recommending `+ "none"` because `module.booleans()` accepts `None` as a valid value. However, `module.booleans()` just returns the unmodified `None` value in that case, so you are correct that it's not necessary. I'm now fine with the code as it is.

PR #252, opened 2017-07-26
Callback option added to junos_jsnapy to control printing of JSNAPy info and error messages, updated jsnapy callback

jnpr-bowen:
Added 'callback' option to junos_jsnapy to control printing of messages set in jsnapy test.
Added code in jsnapy.py to print test info and error messages contained in test
results structure. Whether error and/or info messages are printed is
controlled by the module option 'callback'.

### DIFF
--- a/callback_plugins/jsnapy.py
+++ b/callback_plugins/jsnapy.py
@@ -177,7 +177,11 @@ class CallbackModule(CallbackBase):
                             self._display.display( "\t\t'{0}'".format( str(test['message'])),
                               color=C.COLOR_ERROR)
                             # With jsnapy 1.2 err field parsed into message for each failed test
+                        elif 'xpath_error' in test:
+                          if test['xpath_error'] is True:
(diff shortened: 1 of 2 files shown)


jnpr-bowen:
JSNAPy error messages distinguish between error from test results (error message set), error from XPath error flag set, and general error of test execution.

stacywsmith:
@jnpr-bowen I want to merge/cherry-pick this functionality. However, this depends on JSNAPy 1.2.0. Because of the logging bug in 1.2.0, we really need to wait until the next JSNAPy release before incorporating this functionality. I'll look into this after the next JSNAPy release (which should happen fairly soon).

jnpr-bowen:
That’s fine by me. The only way I’m able to have this work now with the logging bug is to set threads to 1. I have not done a github merge/pull request before, so I could use some pointers on the process.

Later,
Bruce

vnitinv:
@jnpr-bowen Closing this pull request as per discussion above.

PR #256, opened 2017-08-23
Add re name and master state to facts

stacywsmith:


stacywsmith:
Oops. This was already merged under #239.

PR #263, opened 2017-09-18
add junos_pmtud module

mkomon:
Add junos_pmtud module to perform path MTU discovery on Junos devices. 

Finds path MTU to destination IP, with defaults it finds MTU in range 989-1500 B in exactly 9 steps (always log2(max_range) steps). Tolerates any packet loss lower than 100 pct.

### DIFF
--- a/library/junos_pmtud
+++ b/library/junos_pmtud
@@ -74,11 +74,11 @@ options:
     default: None
   dest_ip:
     description:
-        - Destination ip
+        - Destination IPv4 address or hostname
     required: true
   source_ip:
     description:
-        - Source IP used to send the ping
+        - Source IPv4 address used to send the ping
     required: false
   routing_instance:
     description:
@@ -185,6 +185,7 @@ def main():
         msg = 'Unable to connect to {0}: {1}'.format(m_args['host'], str(err))
         module.fail_json(msg=msg)
         return
+
     results = dict(
         changed=False,
         mtu=0,
@@ -205,6 +206,7 @@ def main():
     # Prepare parameters
     ping_params = dict(
             host=m_args['dest_ip'],
+            inet=True,
             count='3',
             do_not_fragment=True,
     )
@@ -230,7 +232,8 @@ def main():
         # Check if ICMP passes
         ping_params['size'] = str(64)
         rpc_reply = dev.rpc.ping(**ping_params)
-        loss = int(rpc_reply.find('probe-results-summary').findtext("packet-loss").strip())
+        loss = int(rpc_reply.findtext('probe-results-summary/packet-loss', default='100').strip())
+
         if loss == 100:
             module.fail_json(
                 msg="Unknown error, PMTUD cannot run.")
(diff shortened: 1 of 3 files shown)


stacywsmith:
@mkomon I like the idea of this module and will review the code today. On a related note...
Would you please consider submitting an additional pull request to add the `routing_instance` argument to `junos_ping`? That would also be helpful.

library/junos_pmtud
```diff
@@ -0,0 +1,263 @@
...
+        changed=False,
+        mtu=0,
+        dest_ip=m_args['dest_ip']
+    )
+
+    # Prepare parameters
+    ping_params = dict(
+            host=m_args['dest_ip'],
+            count='3',
+            do_not_fragment=True,
```
stacywsmith: As it currently stands, a user could enter an IPv6 address for `dest-ip` (or a hostname which resolves to an IPv6 address) and the module will return a value that is 20 bytes less than the actual IPv6 MTU because it assumes IPv4 sized header.

For now, we can force IPv4 only and avoid an incorrect answer by adding `inet=True` to `ping_params`

In the long run, additional arguments and code could be added for other non-IPv4 address families.

library/junos_pmtud
```diff
@@ -0,0 +1,263 @@
...
+        step = int(m_args['max_range'])
+
+        while True:
+            step = step / 2 if step >= 2 else 0
+            ping_params['size'] = str(test_size)
+            rpc_reply = dev.rpc.ping(**ping_params)
+            loss = int(rpc_reply.find('probe-results-summary').findtext("packet-loss").strip())
+            if loss < 100 and test_size == int(m_args['max_size']):
+                # ping success with max test_size, save and break
+                results["mtu"] = test_size
```
stacywsmith: See comment on line 187.

library/junos_pmtud
```diff
@@ -0,0 +1,263 @@
...
+                break
+            elif loss < 100:
+                # ping success, increase test_size
+                results["mtu"] = test_size
```
stacywsmith: See comment on line 187.

library/junos_pmtud
```diff
@@ -0,0 +1,263 @@
...
+    except Exception as err:
+        results['failed'] = True
+        results['msg'] = "unable to execute ping due to:{0}".format(err.message)
+        dev.close()
+        raise err
+
+    if not results["mtu"]:
+        module.fail_json(msg='MTU too low, increase max_range.', **results)
+    else:
+        results["mtu"] += 28    # adjust for IP and ICMP headers
```
stacywsmith: See comment on line 187.

library/junos_pmtud
mkomon: Admittedly, I had only IPv4 in mind when writing this. At the moment I don't have environment with IPv6 running on JUNOS so I agree to limit the module to IPv4 now.

library/junos_pmtud
```diff
@@ -0,0 +1,263 @@
...
+    ## Change default Timeout
+    timeout = int(m_args['timeout'])
+    if timeout > 0:
+        dev.timeout = timeout
+
+    try:
+        # Check if ICMP passes
+        ping_params['size'] = str(64)
+        rpc_reply = dev.rpc.ping(**ping_params)
+        loss = int(rpc_reply.find('probe-results-summary').findtext("packet-loss").strip())
```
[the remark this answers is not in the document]
mkomon: Thanks for the hint! Fixed.

stacywsmith:
@mkomon It looks like you've address all of my feedback except for changing the result key from `mtu` to `inet_mtu`. Let me know if you disagree or have an alternate suggestion.

library/junos_pmtud
mkomon: At first I didn't get this one. Sure, inet_mtu makes sense, I've pushed a new commit.

PR #283, opened 2017-10-23
Rewrite the junos_shutdown and junos_zeroize modules to juniper_junos_system.

stacywsmith:
The new `juniper_junos_system` module replaces both the `junos_shutdown`
and `junos_zeroize` modules. The old modules can still be called and
all arguments should still work for backwards compatibility.

The new module provides takes a required `action` option and provides
some additional capabilities.


action_plugins/_junos_shutdown.py
```diff
@@ -0,0 +1,88 @@
...
+import os.path
+import sys
+
+# The module_utils path must be added to sys.path in order to import
+# juniper_junos_common. The module_utils path is relative to the path of this
+# file.
+module_utils_path = os.path.normpath(os.path.dirname(__file__) +
+                                     '/../module_utils')
+if module_utils_path is not None:
+    sys.path.insert(0, module_utils_path)
```
vnitinv: Why not just use importlib module for this purpose rather adding and deleting path from sys path.

stacywsmith: @vnitinv 
In the case of the action plugin, you are correct, using importlib would work. I will look at switching this code. In the module itself, importlib won't work, and I guess that's why I forgot it would work in the action plugin.

In the action plugin, the module_utils path really is relative to the action plugin's path. In a module, the code has been packaged and placed in a completely different directory and is running with no context of where the original roles directory is located. I believe that has to be handled by manipulating sys.path. See this:
https://github.com/Juniper/ansible-junos-stdlib/blob/master/library/juniper_junos_facts.py#L166-L210

stacywsmith: @vnitinv 

Actually, I don't think importlib will work in the case of the action plugin either. The action plugin is not part of a package. So, even though the code to be imported is in a filename that relative to the action plugin (i.e. ../module_utils/juniper_junos_common.py), it can't be imported with importlib because the action plugin is not part of a package and it can't be made part of a package because of Ansible's role packaging requirements.

FWIW, I tried this code:
```
importlib.import_module('..module_utils.juniper_junos_common', '__name__')
```
and it produced this error:
```
ERROR! Unexpected Exception, this is probably a bug: attempted relative import beyond top-level package
```

PR #305, opened 2017-11-14
update examples to use /project rather than /playbooks

### LINKED ISSUE #304: README references /playbooks as the container path but really it's /project
The README examples say to volume map $PWD to /playbooks. However the WORKDIR is /projects. This is confusing to new users and means you have to specify the full path to the playbook if you follow the example.

Recommends fixing the README to reference the correct path

dgjnpr:
fixes #304


stacywsmith:
@dgjnpr This looks OK to me.

@ntwrkguru Will you please double check that this change looks good to you?

PR #355, opened 2018-04-17
Add support for defining ssh_config path

aristima:
juniper_junos_common.py does not support user defined path for SSH configuration, so by default ~/.ssh/config is used by PyEZ. With this trivial patch one can define alternative ssh_config file, for example /etc/ssh/ssh_config that is then further passed to PyEZ.


vnitinv:
@rsmekala Can you please this till pull request with ssh_key file.

PR #356, opened 2018-04-18
Adding supports_check_mode to import_juniper_junos_common function

rsmekala:


vnitinv:
We should take care of this pull request via #357

PR #357, opened 2018-04-20
Module utils ansiballz integration

rsmekala:
From Ansible 2.1, Ansible uses Ansiballz framework for assembling modules,
But custom module_utils directory is supported from Ansible 2.3 .
Please find the reference for the issue [here](https://groups.google.com/forum/#!topic/ansible-project/J8FL7Z1J1Mw)

Ansiballz packages modules_utils under ansible.module_utils, so juniper_junos_common can be imported by using -
`from ansible.module_utils import juniper_junos_common`

### DIFF
--- a/library/juniper_junos_command.py
+++ b/library/juniper_junos_command.py
@@ -308,7 +308,6 @@ But custom module_utils directory is supported from Ansible 2.3
 Reference for the issue: https://groups.google.com/forum/#!topic/ansible-project/J8FL7Z1J1Mw """
 
 # Ansiballz packages module_utils into ansible.module_utils
-from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils import juniper_junos_common
 
 
--- a/library/juniper_junos_command.py
+++ b/library/juniper_junos_command.py
@@ -303,58 +303,16 @@ stdout_lines:
 import sys
 
 
-def import_juniper_junos_common():
-    """Imports the juniper_junos_common module from _module_utils_path.
+"""From Ansible 2.1, Ansible uses Ansiballz framework for assembling modules
+But custom module_utils directory is supported from Ansible 2.3
+Reference for the issue: https://groups.google.com/forum/#!topic/ansible-project/J8FL7Z1J1Mw """
 
-    Ansible versions < 2.4 do not provide a way to package common code in a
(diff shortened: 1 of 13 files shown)


stacywsmith:
What problem are you trying to solve? Other than the `supports_check_mode` issue, which has already been resolved, is there another issue that needs fixing?

I readily admit the `import_juniper_junos_common()` function is a hack, but it was a purposeful and intentional hack because we wanted to maintain backwards compatibility with the oldest version of Ansible possible in order to maximize utility for users. I understand versions < 2.4 are not supported by Ansible, but my experience is that a significant number of users are still using these older versions.

If removing this hack and only supporting Ansible >= 2.3 actually provides some benefit, I'm supportive of the change. Otherwise, I'd recommend waiting to remove this hack until there's a real need/benefit.

library/juniper_junos_command.py
```diff
@@ -303,57 +303,16 @@
...
-        bypass_checks=True
-    )
-    import_path = module.params.get('_module_utils_path')
-    if import_path is not None:
-        sys.path.insert(0, import_path)
-        import juniper_junos_common
-        del sys.path[0]
-    return juniper_junos_common
+# Ansiballz packages module_utils into ansible.module_utils
+from ansible.module_utils.basic import AnsibleModule
```
stacywsmith: Can you explain why this import is necessary? It doesn't appear to be used.

rsmekala:
@stacywsmith , As you said Ansible < 2.3 are not officially supported by Ansible.  We had asked in the  ATT forum for any users currently running Ansible < 2.3, but we didn't get any reply. So we assumed that not many users use Ansible < 2.3. It was also recommended by Ansible group to drop support for 
    Ansible < 2.3. Please find the reference to the [issue](https://groups.google.com/forum/#!topic/ansible-project/J8FL7Z1J1Mw) here.

If some user prefers Ansible < 2.3 , we will suggest them to use upto 2.0.2 version hosted on ansible-galaxy.

Ansible was designed for a remote-tier architecture i.e playbooks can be executed on remote machines. Using `import_juniper_junos_common` function to find juniper_junos_common module would not be possible on the remote machine. Using Ansiballz framework will help us package the module_utils into the .tar file sent to remote machine via scp. This feature will be useful for upcoming Junos flex.

Issue #354, opened 2018-04-17
logging changes between 1.4 and 2.0


fooelisa:
Hi, after an upgrade from 1.4 and 2.0 I've noticed some changes in the logging behavior. I am not sure about the feature to configure the log level though Ansible's verbosity. I'd like for the router log file to be more chatty than ansible and what the user has to see on the CLI. I can't depend on the user to always remember to attach the required number of v's and there doesn't seem to be a way of permanently setting the log level for ansible at the moment either. Did you guys have anything in mind for such a scenario? 

Thanks
Elisa

stacywsmith:
In my opinion (remembering I no longer speak for Juniper) this is a reasonable and worthwhile request.

I believe this could be implemented fairly easily with:
1) Specify an additional logging-specific option, such as `level`, [here](https://github.com/Juniper/ansible-junos-stdlib/blob/master/module_utils/juniper_junos_common.py#L497-L501) and [here](https://github.com/Juniper/ansible-junos-stdlib/blob/master/module_utils/juniper_junos_common.py#L323-L384)
I would suggest that the `level` argument could use Ansible's `choices` functionality to restrict the value to only known logging levels.
 
2) Use the value of the new option to set the logging level [here](https://github.com/Juniper/ansible-junos-stdlib/blob/master/module_utils/juniper_junos_common.py#L868-L876).

vnitinv:
These were some changes done in current github code https://github.com/Juniper/ansible-junos-stdlib/pull/337/files which is not present in galaxy. 
@fooelisa from were did you install roles? So if you install from github and provide logfile option, all the roles related logs should go in that given logfile.

fooelisa:
Got the latest github running, not using galaxy. They go into the logfile alright, configuring the verbosity is the issue.

rsmekala:
@fooelisa Could you try the following command:

`export ANSIBLE_DEBUG=True`

This will set the logging level of both Ansible and Juniper.junos module to DEBUG level. The fix has been mentioned in [Ansible Docs](https://docs.ansible.com/ansible/2.4/network_debug_troubleshooting.html#enabling-networking-logging-and-how-to-read-the-logfile)

Let us know the result.

fooelisa:
[...]

Issue #340, opened 2018-03-14
juniper_junos_config returns "check_mode not supported" when attempting to run ansible-playbook --check


itsgc:
Good Morning,

We have a mostly juniper fleet, with 100% ansible coverage on access switches and various degree of coverage on the Layer 3 gear. We are running version 1.4.2 of ansible-junos-stdlib.

Currently, we tipically run ansible in check_mode first, validate the diff that gets produced and if deemed acceptable, the neteng re-runs ansible without check_mode to deploy the change and commit it to running config.

I ran into a particularly troublesome issue for us with regards to check_mode after trying to upgrade to 2.0.2.

Upgrading the lib to 2.0.2 yields errors about check_mode not being supported by juniper_junos_config.

Updating our playbook to use the new module arguments does not change the result.

Only forcibly disabling check_mode for the task and applying some jinja2 magic to replicate the old check_mode behavior obtains the intended workflow but we don't see this as a viable alternative.

I was hoping to confirm if check_mode is intentionally unsupported or this might be a bug or maybe a misconfiguration on my part?

Thanks for any help or insight you might provide.

**Please find the details of the issue attached here:**

Our Current Setup that works as intended:
Ansible 2.3.1.0
ansible-junos-stdlib 1.4.2

My test environment:
Ansible 2.3.1.0
ansible-junos-stdlib 2.0.2

Command line used to launch ansible: 
ansible-playbook -i prod.hosts site.yml -e user=myusername --check -f25 --ask-vault-pass --limit DEVICE

Relevant snippet of our current playbook (**functional with 1.4.2**):
[playbook_snippet_1.4.2.txt](https://github.com/Juniper/ansible-junos-stdlib/files/1813163/playbook_snippet_1.4.2.txt)

Output of the task:
`TASK [Load the compiled configuration via "load replace"]`
`*************************************************************`
`changed: [DEVICE] => {"changed": true, "file": "PATH/compiled/DEVICE/device.conf"}`

Upgrading ansible-junos-stdlib to 2.0.2 without changing the playbook or anything else yields:

`TASK [Load the compiled configuration via "load replace"]` `*************************************************************`
`skipping: [DEVICE] => {"changed": false, "msg": "remote module (junos_install_config) does not support check mode", "skipped": true}`

Revised playbook using 2.0.2. (**non functional**) - this uses "modern" module names and arguments however it does not change the end result:
[...]

PR #359, opened 2018-04-26
Logging level

rsmekala:
Fix for issue #354 
Added level option to logging specifications. User can use `level` option in task to set the verbosity level for Juniper.junos role.

User can choose of either INFO/DEBUG logging level. The default level would be WARNING.

The order of preference for setting the level would be:
1. Set level to DEBUG if ANSIBLE_DEBUG is True.
2. Set level based on number of v's passed to module.
3. Set level based on level option mentioned in task data.

### DIFF
--- a/module_utils/juniper_junos_common.py
+++ b/module_utils/juniper_junos_common.py
@@ -381,6 +381,20 @@ class ModuleDocFragment(object):
         type: path
         aliases:
           - log_file
+      level:
+        description:
+          - The level of information to be logged can be modified using this option
+          - 1) By default, messages at level C(WARNING) or higher are logged.
+          - 2) If the C(-v) or C(--verbose) command-line options to the
+               C(ansible-playbook) command are specified, messages at level
+               C(INFO) or higher are logged.
+          - 3) If the C(-vv) (or more verbose) command-line option to the
+               C(ansible-playbook) command is specified, or the C(ANSIBLE_DEBUG)
+               environment variable is set, then messages at level C(DEBUG) or
+               higher are logged.
+          - 4) If I(level) is mentioned then messages at level I(level) or more are
+               logged.
+
 '''
 
     # _SUB_CONNECT_DOCUMENTATION is just _CONNECT_DOCUMENTATION with each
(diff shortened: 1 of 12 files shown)


module_utils/juniper_junos_common.py
```diff
@@ -497,12 +511,15 @@ class ModuleDocFragment(object):
...
-    'logdir': dict(type='path', required=False, default=None)
+    'logdir': dict(type='path', required=False, default=None),
+    'level': dict(choices=[None, 'INFO', 'DEBUG'], required=False, default=None)
 }
 
 # The logdir and logfile options are mutually exclusive.
 logging_spec_mutually_exclusive = ['logfile', 'logdir']
 
+
+
```
stacywsmith: Why add extra unnecessary whitespace? Suggest you always run `pep8` against your changes.

module_utils/juniper_junos_common.py
```diff
@@ -652,7 +669,7 @@ def __init__(self,
         # Update argument_spec with the top_spec
         argument_spec.update(top_spec)
         # Extend mutually_exclusive with connection_mutually_exclusive
-        mutually_exclusive += top_spec_mutually_exclusive
+        mutually_exclusive = top_spec_mutually_exclusive
```
stacywsmith: This will break things! A module may specify parameters which are mutually exclusive. The common parameters which are mutually exclusive should be **added** to the module specified parameters, not override them.

vnitinv:
@rsmekala Can you please clean up unwanted commits

stacywsmith:
@rsmekala you added a commit that reverts the `supports_check_mode` fix, rather than updating your sandbox to include this commit. Now, if you were to merge this PR as it currently stands, you would actually revert the previous fix.

vnitinv:
Fix #359

rsmekala:
@stacywsmith The pull request adding supports_check_module to the role is still not merged (#356 )

We are planning to go with #357 to address the issue #340.

vnitinv:
@rsmekala This pull request is clean as of now with respect to other changes in the discussion. So I think we are good.

PR #366, opened 2018-05-15
Dockerfile update to include pip

rsmekala:
Made changes required to install pip in juniper/pyez-ansible container.
- apk update to avoid pip installation errors.
- Install pip from py-pip

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ ADD version.py /tmp/ansible-junos-stdlib/version.py
 
 
 RUN tar -czf Juniper.junos ansible-junos-stdlib &&\
+    apk add python2-dev &&\
     apk update && apk add ca-certificates &&\
     apk add openssh-client &&\
     apk add build-base gcc g++ make python-dev &&\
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,6 +24,7 @@ RUN tar -czf Juniper.junos ansible-junos-stdlib &&\
     apk update && apk add ca-certificates &&\
     apk add openssh-client &&\
     apk add build-base gcc g++ make python-dev &&\
+    apk update && apk add py-pip &&\
```
vnitinv: apk update is already there in line 24. Why to call this twice. Please check.

rsmekala: It gives the following error, if I try to install py-pip without `apk update`:

```
WARNING: Ignoring APKINDEX.84815163.tar.gz: No such file or directory
WARNING: Ignoring APKINDEX.24d64ab1.tar.gz: No such file or directory
ERROR: unsatisfiable constraints:
  py-pip (missing):
    required by: world[py-pip]
```

I had found the above fix [here](https://github.com/gliderlabs/docker-alpine/issues/207).

PR #367, opened 2018-05-15
Fixed junos_jsnapy callback plugin

rsmekala:
- From Ansible 2.4, module_name is not passed as part of result to
 v2_runner_on_ok(). So, the current code returns a Null object without
 adding the hosts.
- Corrected by removing usage of module_name to add hosts in
 junos_jsnapy callback plugin.

### DIFF
--- a/callback_plugins/jsnapy.py
+++ b/callback_plugins/jsnapy.py
@@ -31,25 +31,20 @@ class CallbackModule(CallbackBase):
 
   def v2_runner_on_ok(self, result):
     """
-    Collect test results for all tests executed if module is junos_jsnapy
+    Collect test results for all tests executed if action is snapcheck or check
     """
 
     ## Extract module name
-    module_name = ''
     module_args = {}
     if 'invocation' in result._result:
-      if 'module_name' in result._result['invocation']:
-        module_name = result._result['invocation']['module_name']
+      if 'module_args' in result._result['invocation']:
         module_args = result._result['invocation']['module_args']
 
     ## Check if dic return has all valid information
-    if module_name == '' or module_args == {}:
-        return None
-    elif 'action' not in module_args:
+    if 'action' not in module_args:
         return None
 
-    if module_name == 'junos_jsnapy' and \
-    ( module_args['action'] == 'snapcheck' or module_args['action'] == 'check' ):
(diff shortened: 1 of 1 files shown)


vnitinv:
`From Ansible 2.4, module_name is not passed `

@rsmekala Is this documented anywhere in Ansible?

rsmekala:
I don't think, it is documented. But, there is a [issue](https://github.com/ansible/ansible/issues/27398) for this.

The current `result._result['invocation']` would contain:
```
'invocation': {
    u 'module_args': {
      u 'baud': None,
      u 'attempts': None,
      u 'logfile': u 'ansible.log',
      u 'config_file': None,
      u 'passwd': u 'VALUE_SPECIFIED_IN_NO_LOG_PARAMETER',
      u 'logdir': None,
      u 'ssh_config': None,
      u 'host': u 'xxxx',
      u 'user': u 'xxxx',
      u 'timeout': 30,
      u 'provider': None,
      u 'action': u 'snapcheck',
      u 'console': None,
      u 'ssh_private_key_file': None,
      u 'test_files': [u 'tests/junos_jsnapy/test_junos_storage.yml'],
      u 'port': xx,
      u 'dir': u '/etc/jsnapy/testfiles',
      u 'mode': None
    }
}
```

I had also searched for any [Ansible core callback plugins](https://github.com/ansible/ansible/tree/devel/lib/ansible/plugins/callback) using module_name, but couldn't find any. Most of the selective callback plugins decide their action based on `module_args`

jnpr-bowen:
A quick review of this fix show it does not limit the data collected to the juniper_junos_jsnapy module. The 'if' condition on the module_name has been removed and so the behavior with only checking the module_args['action'] could be wrong. From my own reading of the ansible core callback plugins/callbacks/mail.py (line 168), it looks like the module_name equivalent has moved to result._task.action:
`

         body = 'Playbook: %s\n' % os.path.basename(self.playbook._file_name)

         if result._task.name:

             body += 'Task: %s\n' % result._task.name

         body += 'Module: %s\n' % result._task.action

         body += 'Host: %s\n' % host

         body += '\n'

`
That is what I am working from on my callback that will present the info and error messages from the jsnapy tests.

PR #370, opened 2018-05-18
[WIP] New module juniper_junos_file

dgarros:
I've started to work on a new module to manage file, it's based on @renetbo work in #177 and I tried to refactor to align with the new structure.
I've been using it to transfet file to/from junos devices and this part is working good so far.

I still need to do more testing but I've seen some discussions around a module like this one so I figured I would create the PR for visibility.

### DIFF
--- a/action_plugins/juniper_junos_file.py
+++ b/action_plugins/juniper_junos_file.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+
+#
+# Copyright (c) 2017-2018, Juniper Networks Inc. All rights reserved.
+#
+# License: Apache 2.0
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright
+#   notice, this list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright
+#   notice, this list of conditions and the following disclaimer in the
+#   documentation and/or other materials provided with the distribution.
+#
+# * Neither the name of the Juniper Networks nor the
+#   names of its contributors may be used to endorse or promote products
+#   derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY Juniper Networks, Inc. ''AS IS'' AND ANY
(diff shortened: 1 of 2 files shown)


action_plugins/juniper_junos_file.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
```
stacywsmith: I think this action plugin is the same as juniper_junos_common_action.py. If it is, it should be a link to juniper_junos_common_action.py rather than a copy/paste of the file. The diff doesn't make it clear if this is a file or a link, so if it's already a link, please ignore my comment.

library/juniper_junos_file.py
```diff
@@ -0,0 +1,336 @@
...
+
+from __future__ import absolute_import, division, print_function
+import hashlib
+from jnpr.junos.utils.scp import SCP
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'supported_by': 'community',
+                    'status': ['stableinterface']}
+
+DOCUMENTATION = '''
```
stacywsmith: Please make sure the documentation for this module builds correctly before we merge. The documentation syntax is very fragile.

library/juniper_junos_file.py
```diff
@@ -0,0 +1,336 @@
...
+changed:
+  description:
+    - Indicates if the device's state has changed.
+  returned: when the file has been successfully copied.
+  type: bool
+
+'''
+
+
+def import_juniper_junos_common():
```
stacywsmith: This function is no longer necessary after #357 and should be removed.

library/juniper_junos_file.py
```diff
@@ -0,0 +1,336 @@
...
+       else:
+           raise err
+       return checksum
+    junos_module.logger.info("rpc reponse recvd")
+    return checksum
+
+
+def main():
+    # Import juniper_junos_common
+    juniper_junos_common = import_juniper_junos_common()
```
stacywsmith: Remove.

jnpr-community-netdev:

jnpr-community-netdev:

dineshbaburam91:
Fixed in collection

Issue #322, opened 2018-01-12
No module named plugins.action.normal using 2.0.0


hl1234:
I have found myself in a confusing situation since I just started trying to use Juniiper.junos this week with most of my progress yesterday.  I was also going through rebuilding my python environment with pyenv.  So I got back to a working set of playbooks and decided to try the 2.0.0 version.  That is when I encountered the error "No module named plugins.action.normal". The only change I made between work and not working was to remove 1.4.3 and install 2.0.0.

This may be an obvious newbie problem or something real so I will defer to you wrt prioritization. In the meantime I will go back to 1.4.3 and check back later.

I use -p with galaxy to install Juniper.junos in the local roles/ directory where I am working.

More follows

---------
$ ansible --version
ansible 2.4.2.0
  config file = /home/harry/Projects/network-configuration/ansible/ansible.cfg
  configured module search path = [u'/home/harry/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/harry/.pyenv/versions/2.7.14/lib/python2.7/site-packages/ansible
  executable location = /home/harry/.pyenv/versions/2.7.14/bin/ansible
  python version = 2.7.14 (default, Jan 12 2018, 10:31:32) [GCC 4.8.4]
$ python --version
Python 2.7.14

------

$ pip list
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
asn1crypto (0.24.0)
bcrypt (3.1.4)
cffi (1.11.2)
cryptography (2.1.4)
enum34 (1.1.6)
idna (2.6)
ipaddress (1.0.19)
Jinja2 (2.10)
junos-eznc (2.1.7)
jxmlease (1.0.1)
lxml (4.1.1)
MarkupSafe (1.0)
ncclient (0.5.3)
netaddr (0.7.19)
paramiko (2.4.0)
pip (9.0.1)
pyasn1 (0.4.2)
pycparser (2.18)
PyNaCl (1.2.1)
pyserial (3.4)
PyYAML (3.12)
scp (0.10.2)
setuptools (38.4.0)
six (1.11.0)
wheel (0.30.0)

-------

The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_NvO_Cq/ansible_module_junos_get_config.py", line 1164, in <module>
    main()
[...]

PR #388, opened 2018-08-09
Change hardcoded /usr/bin/python 

funzoneq:
to /usr/bin/env python to support virtualenv. It is really irritating to have to specify an ansible_python_interpreter when this is just build into your OS by default.

Ansible does the same: https://github.com/ansible/ansible/blob/devel/bin/ansible#L1

Addresses #322

### DIFF
--- a/library/juniper_junos_command.py
+++ b/library/juniper_junos_command.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # Copyright (c) 1999-2018, Juniper Networks Inc.
--- a/library/juniper_junos_config.py
+++ b/library/juniper_junos_config.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # Copyright (c) 1999-2018, Juniper Networks Inc.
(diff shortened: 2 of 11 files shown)


funzoneq:
Can this get merged? @jeremyschulman @vnitinv @rsmekala

stacywsmith:
Have you tested this change?

Please see my [comment](https://github.com/Juniper/ansible-junos-stdlib/issues/322#issuecomment-368062172) in #322.

Specifically:
> In the absence of a ansible_python_interpreter setting, older versions of Ansible would execute with the interpreter specified in the module's shebang line, but newer versions of Ansible completely ignore the shebang line.

I admittedly have not been working with Ansible for the last 8 months, so it's possible Ansible's behavior on this has changed within that time.

stacywsmith:
> Ansible does the same: https://github.com/ansible/ansible/blob/devel/bin/ansible#L1

This is not comparable. This link is for the actual `ansible` CLI command, not a module.

Modules execute in a separate process and are invoked by Ansible, not the shell.

stacywsmith:
Also see the Ansible documentation for [ansible_python_interpreter](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#ansible-python-interpreter).

stacywsmith:
Also see this [must statement](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#python-shebang) from the Ansible module developer's guide.

PR #418, opened 2019-02-28
Add support for Console over SSH (finally)

dgarros:
After some time we are finally getting back to this topic.

This PR is adding support for Console over SSH to all modules by adding support for 2 new parameters `cs_user` and `cs_passwd`

The way the support for Console over SSH has been implemented in Pyez, the option `mode` is not really required so I was not sure if we should had something specific or not.


jnpr-community-netdev:

jnpr-community-netdev:

vnitinv:
@rsmekala @dineshbaburam91 can we cherry-pick the changes. We don't need the changes done in requirements.tzt

rsmekala:
@dgarros Thanks for the contribution. Merging this (finally 😄 )

PR #428, opened 2019-05-01
Add support for ignore_warning in juniper_junos_common/open_configuration

dgarros:
I noticed that `juniper_junos_common/open_configuration` wasn't respecting the `ignore_warning` parameter.

In our environment we were getting some error when 2 tasks are trying to run a config diff + check at the same time (even in private mode)
```
fatal: [rs1-aw38-c1-ash1 -> localhost]: FAILED! => changed=false
  msg: |-
    Unable to open the configuration in private mode: RpcError(severity: warning, bad_element: None, message: warning: changes cannot be committed while 'configure exclusive' is active
    warning: uncommitted changes will be discarded on exit)
```
Here is the task
```yaml
     juniper_junos_config:
        retrieve: candidate
        load: overwrite
        config_mode: private
        diff: true
        check: true
        commit: false
        src: "{{ junos_conf }}"
        provider: "{{ credential }}"
```
> there might be another issue here because we shouldn't have a lock on the configuration at this time since all tasks are running in `private` mode but I couldn't find where this lock was coming from so I fixed the ignore_arning issue... 

Looking at the code, this parameter was not passed to the function but a default value was already defined for `rpc.open_configuration`
```python
self.dev.rpc.open_configuration(
                        private=True,	                    
                        ignore_warning='uncommitted changes will be '	                       
                                       'discarded on exit')
```

This patch passed the `ignore_warning` arg to `open_configuration` and follow this logic: 
- if `ignore_warning` is a bool, pass the bool
- if `ignore_warning` is a string, create a list and add it to the default value
- if `ignore_warning` is a list, add the default value to the list

### DIFF
--- a/library/juniper_junos_config.py
+++ b/library/juniper_junos_config.py
@@ -1001,7 +1001,7 @@ def main():
 
     junos_module.logger.debug("Step 1 - Open a candidate configuration "
                               "database.")
-    junos_module.open_configuration(mode=config_mode)
+    junos_module.open_configuration(mode=config_mode, ignore_warning=ignore_warning)
     results['msg'] += 'opened'
 
     junos_module.logger.debug("Step 2 - Load configuration data into the "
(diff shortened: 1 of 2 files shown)


jnpr-community-netdev:

jnpr-community-netdev:

rsmekala:
ok to test

rsmekala:
retest this please

rsmekala:
retest this please

Issue #434, opened 2019-05-21
get-config rpc returns empty when format="text"


dineshbaburam91:
Issue Type
------
 - Bug Report

Module Name
------
```
$ pip list
Package      Version 
------------ --------
ansible      2.4.2.0 
asn1crypto   0.24.0  
bcrypt       3.1.4   
certifi      2019.3.9
cffi         1.11.5  
chardet      3.0.4   
colorama     0.4.1   
configparser 3.7.4   
cryptography 2.2.2   
enum34       1.1.6   
future       0.17.1  
icdiff       1.9.1   
idna         2.6     
ipaddress    1.0.22  
Jinja2       2.10    
jsnapy       1.3.2   
junos-eznc   2.2.1   
jxmlease     1.0.1   
lxml         4.2.1   
MarkupSafe   1.0     
ncclient     0.6.4   
netaddr      0.7.19  
paramiko     2.4.1   
pip          10.0.1  
pyasn1       0.4.2   
pycparser    2.18    
PyNaCl       1.2.1   
pyparsing    2.4.0   
pyserial     3.4     
PyYAML       5.1     
requests     2.21.0  
scp          0.10.2  
selectors2   2.0.1   
setuptools   39.1.0  
six          1.11.0  
urllib3      1.24.2  
wget         3.2     
wheel        0.31.0  

$ ansible --version
ansible 2.4.2.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/Users/babud/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/babud/automation_development/ansible_junos/ansiblevir/lib/python2.7/site-packages/ansible
  executable location = /Users/babud/automation_development/ansible_junos/ansiblevir/bin/ansible
  python version = 2.7.10 (default, Feb 22 2019, 21:17:52) [GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.37.14)]
```

OS / Environment
------
All platform

Summary
------
get-config rpc returns empty when format="text"

Steps to reproduce
------

```
  tasks:
    - name: Get system information
      juniper_junos_rpc:
        rpcs: "get-config"
        format: "text"
      register: response
```

Expected results
```
Should return the output of get_config rpc
```
Actual results
------

```
ok: [x.x.x.x] => {
    "response": {
        "attrs": null, 
        "changed": false, 
        "failed": false, 
        "format": "text", 
[...]

PR #435, opened 2019-05-22
Fixes issue #434

rsmekala:
RPC: `get_config` is treated specially in both PyEZ and juniper_junos_rpc. The current code doesn't pass the format option to PyEZ's `get_config`, which returns an XML response instead of text response, causing the incorrect behavior.

Corrected the behavior by passing down the format option to PyEZ.

Links:

https://github.com/Juniper/py-junos-eznc/blob/472408c3e7a8dd67001c09589c0cf35faad7e829/lib/jnpr/junos/rpcmeta.py#L26-L38

### DIFF
--- a/library/juniper_junos_rpc.py
+++ b/library/juniper_junos_rpc.py
@@ -510,6 +510,8 @@ def main():
                     attr = {}
                 if kwarg is None:
                     kwarg = {}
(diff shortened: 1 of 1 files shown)


rsmekala:
retest this please

PR #438, opened 2019-06-19
Adding docs to clarify the usage of ssh_private_key_file

ntwrkguru:
In newer versions of OpenSSH, keys are generated in the OPENSSH format, but paramiko needs the RSA PEM format. I've added instructions to check for, convert, and create keys in the proper format.


jnpr-community-netdev:

jnpr-community-netdev:

PR #427, opened 2019-04-18
--diff is not working correctly

marioschaefer:
--diff doesn't show the diff as expected.
we changed the handling for the `juniper_junos_config` a bit to return the diff correctly, so that ansible is showing it.

output before the fix:
```
ansible-playbook sitegroups/bgp-router.yml -l XYZ -CD

PLAY [provision bgp-routers] *************************************

TASK [bgp-router : install system configuration] *************************************

changed: [XYZ]

PLAY RECAP *************************************
XYZ            : ok=4    changed=1    unreachable=0    failed=0
```

output after the change:
```
ansible-playbook sitegroups/bgp-router.yml -l XYZ -CD

PLAY [provision bgp-routers] *************************************

TASK [bgp-router : install system configuration] *************************************

[edit system root-authentication]
-   encrypted-password "asdfasdfasdf."; ## SECRET-DATA
+   encrypted-password "ddddddddd."; ## SECRET-DATA
changed: [XYZ]

PLAY RECAP *************************************
XYZ           : ok=4    changed=1    unreachable=0    failed=0
```

PR #462, opened 2019-09-23
juniper_junos_config: update doc

jacek-izykowski:
Change #427 modified behavior of juniper_junos_config (diff is not
a string anymore - it's a dict now). Behavior is now inconsistent
with the documentation.


jnpr-community-netdev:

rsmekala:
ok to test

PR #466, opened 2019-11-13
Adjusted priority between self.params and self.params['provider']

rsmekala:
Some parameters like `logfile` and `logdir` can either be in `provider` or declared in the module params. The priority should be given to the module params in this case.

I have verified the fix for the following scenarios using `logdir` :
- Declared in module params and not in `provider`.
- Declared in `provider` and not in module params.
- Declared in both provider and module. It correctly picks up the directory in module params.


rsmekala:
ok to test

rsmekala:
retest this please

rahkumar651991:
Providers are not supported on collection so this will not be merged. Closing it.

PR #469, opened 2019-12-11
Added tests for juniper_junos_rpc

rsmekala:
I have added the test in Jenkins as well.

Thanks to the tests. I was able to find some minor bugs in `juniper_junos_common` 

- In case we are not able to convert it to bool, we should return `None`. Currently, it causes a `TypeError`.
- The mode should be `ab` instead of `a` to allow python3 to write bytestrings

### DIFF
--- a/module_utils/juniper_junos_common.py
+++ b/module_utils/juniper_junos_common.py
@@ -1177,7 +1177,6 @@ class JuniperJunosModule(AnsibleModule):
                         bool_val = boolean(v)
                     except TypeError:
                         bool_val = None
-                        pass
                     if bool_val is not None:
                         v = bool_val
                 return_item[k] = v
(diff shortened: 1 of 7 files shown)


rsmekala:
ok to test

rsmekala:
retest this please

module_utils/juniper_junos_common.py
```diff
@@ -1173,7 +1173,11 @@ def parse_arg_to_list_of_dicts(self,
                 if allow_bool_values is True:
                     # Try to convert it to a boolean value. Will be None if it
                     # can't be converted.
-                    bool_val = boolean(v)
+                    try:
+                        bool_val = boolean(v)
+                    except TypeError:
+                        bool_val = None
+                        pass
```
vnitinv: remove this line

rsmekala:
retest this please

module_utils/juniper_junos_common.py
rsmekala: Fixed

PR #470, opened 2019-12-11
Updated dockerfile to install Juniper.junos via galaxy

rsmekala:
Logs:

```
/playbooks # ansible-galaxy list
# /root/.ansible/roles
- Juniper.junos, 2.2.1
[WARNING]: - the configured path /usr/share/ansible/roles does not exist.

[WARNING]: - the configured path /etc/ansible/roles does not exist.

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,22 +4,18 @@ LABEL net.juniper.image.maintainer="Stephen Steiner <ssteiner@juniper.net>"
 
 WORKDIR /tmp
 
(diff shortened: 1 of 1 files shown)


ntwrkguru:
Those warnings are OK as the role is installed into a different place, but usable just the same. Installing them from galaxy would mean that one cannot test changes made to an unpublished version of the role. This is now negated with the move to collections, so we can probably close it, I think. @rahkumar651991 ?

rahkumar651991:
closing this as the changes are not relevant.

Issue #379, opened 2018-06-19
juniper_junos_table fails if subtable contains no elements


ferfemar:
Issue Type
------

 - Bug Report

Module Name
------

juniper_junos_table

Juniper.Junos role and Python libraries version

```
$ ansible --version
ansible 2.5.2
  config file = /home/ferfecky/ansible_playground/detailed-interfaces/ansible.cfg
  configured module search path = [u'/home/ferfecky/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.13 (default, Nov 24 2017, 17:33:09) [GCC 6.3.0 20170516]

$ 

$ pip freeze
ansible==2.5.2
appdirs==1.4.3
asn1crypto==0.24.0
attrs==16.3.0
bcrypt==3.1.4
beautifulsoup4==4.5.3
certifi==2018.4.16
cffi==1.11.5
chainmap==1.0.2
chardet==3.0.4
click==6.6
colorama==0.3.9
configparser==3.5.0
constantly==15.1.0
cryptography==2.2.2
enum34==1.1.6
future==0.16.0
gpg==1.8.0
html5lib==0.999999999
httplib2==0.9.2
icdiff==1.9.1
idna==2.7
incremental==16.10.1
ipaddress==1.0.22
Jinja2==2.10
jsnapy==1.3.1
jtextfsm==0.3.1
junos-eznc==2.1.8
jxmlease==1.0.1
keyring==10.1
keyrings.alt==1.3
librouteros==1.0.4
lxml==4.2.1
M2Crypto==0.24.0
MarkupSafe==1.0
napalm==2.3.1
napalm-ansible==0.9.1
napalm-base==0.25.0
napalm-eos==0.6.1
napalm-fortios==0.4.0
napalm-ios==0.8.1
napalm-iosxr==0.5.6
napalm-junos==0.12.1
napalm-nxos==0.7.1
napalm-panos==0.4.0
napalm-pluribus==0.5.1
napalm-ros==0.2.2
napalm-vyos==0.1.3
ncclient==0.5.3
netaddr==0.7.19
netmiko==2.1.1
netsnmp-python==1.0a1
packaging==16.8
PAM==0.4.2
pan-python==0.12.0
paramiko==2.4.1
pbr==3.1.1
ply==3.10
pyasn1==0.4.3
[...]

PR #485, opened 2020-02-19
Update juniper_junos_table.py

kr3ator:
Checking for bool(value) when it is a nested table with no entries returns False, hence passes the Table object as a value instead of an empty list. In the end this fails in Ansible becuase it doesn't know how to digest that into a text.

This fixes issue #379


jnpr-community-netdev:

PR #488, opened 2020-03-16
fix for ignore_warn_list for string types

rahkumar651991:
Fix added if ignore_warn_list has string type character elements.


jnpr-community-netdev:

PR #496, opened 2020-04-29
Enabling whitelist feature for jsnapy callback

rahkumar651991:
Adding whitelist to be added in ansible.cfg for jsnapy callback to be enabled or disabled.
jsnapy callback will no more be enabled by default.

### DIFF
--- a/callback_plugins/jsnapy.py
+++ b/callback_plugins/jsnapy.py
@@ -20,6 +20,9 @@ class CallbackModule(CallbackBase):
   CALLBACK_TYPE = 'aggregate'
   CALLBACK_NAME = 'jsnapy'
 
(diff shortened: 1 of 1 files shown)


callback_plugins/jsnapy.py
```diff
@@ -20,6 +20,9 @@ class CallbackModule(CallbackBase):
   CALLBACK_TYPE = 'aggregate'
   CALLBACK_NAME = 'jsnapy'
 
+  # only needed if you ship it and don't want to enable by default
```
vnitinv, line 23: In what context "only needed if you ship"? we are shipping callback so its not optional.

rahkumar651991: modified the comment.

PR #499, opened 2020-05-07
Adding ignore warning for diff in juniper junos config

rahkumar651991:
Pass and parse the ignore_warning argument in diff for juniper_junos_config.

### DIFF
--- a/module_utils/juniper_junos_common.py
+++ b/module_utils/juniper_junos_common.py
@@ -1570,7 +1570,7 @@ class JuniperJunosModule(AnsibleModule):
 
         self.logger.debug("Diffing candidate and committed configurations.")
         try:
-	    rb_id =0
+            rb_id =0
             diff = self.config.diff(rb_id, ignore_warning)
             self.logger.debug("Configuration diff completed.")
             return diff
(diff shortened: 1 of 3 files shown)


vnitinv:
@rahkumar651991 @dineshbaburam91  Why our jenkins didnt run here? Can you please check this

module_utils/juniper_junos_common.py
```diff
@@ -1570,7 +1570,8 @@ def diff_configuration(self):
 
         self.logger.debug("Diffing candidate and committed configurations.")
         try:
-            diff = self.config.diff(rb_id=0)
```
vnitinv, line 1573: why you had to split it into 2 lines,
we can just say
```python
self.config.diff(rb_id=0, ignore_warning=ignore_warning)
```

vnitinv:
why jenkins is not running @rahkumar651991

vnitinv:
ok to test

zephyr2095:
I believe this change broke something.  When I try to push config I get this error:

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: diff() got an unexpected keyword argument 'ignore_warning'

If I remove the changes made, the module works correctly.

rahkumar651991:
@zephyr2095 - It seems you will need to upgrade junos-eznc(PyEZ). It has related changes in there also. 
If that doesn't work share the version details for ansible-roles and junos-eznc.

zephyr2095:
@rahkumar651991  - updating junos-eznc to 2.5.0 fixed it!  Thanks for the help!

PR #512, opened 2020-07-09
Stop passing ignore_warning param to diff()

bcavns01:
diff() function in the required junos-eznc version (>=2.2.0) no longer has ignore_warning parameter.

### DIFF
--- a/module_utils/juniper_junos_common.py
+++ b/module_utils/juniper_junos_common.py
@@ -1580,7 +1580,7 @@ class JuniperJunosModule(AnsibleModule):
 
         self.logger.debug("Diffing candidate and committed configurations.")
         try:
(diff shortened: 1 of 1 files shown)


rahkumar651991:
https://github.com/Juniper/ansible-junos-stdlib/pull/499 ---> Refer this. 
Ignore-warning has recently been added. This pull request is not valid.
Would recommend to use juniper.junos 2.4.0 (ansible roles) and junos-eznc(2.5.0) to stay in sync.

bcavns01:
In that case, it seems `MIN_PYEZ_VERSION = "2.2.0"` need to be changed here instead.

rahkumar651991:
yes, We modified the pyez version in requirements page. 
Need to be done there also. Thanks for pointing it out.

PR #516, opened 2020-07-23
change legacy /project to /playbooks in documentation

stmorim:


ntwrkguru:
The volume declaration is still `/project` to align with the concept that it's not just a playbook(s), but likely also roles, modules, variables, and configuration. This is mirrored in AWX/Tower where the term project is also used.

Dockerfile, line 30:

```
VOLUME    /project
```

rahkumar651991:
closing the PR based on the comment from @ntwrkguru .

Issue #589, opened 2022-07-15
Incompatibility with Ansible 4+ due to API change


TheMysteriousX:
Issue Type
------

 - Bug Report

Module Name
------

juniper.device.rpc

juniper.device collection and Python libraries version
------

```
$ ansible --version
ansible [core 2.13.1]
  config file = /builds/operations/network/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.9.10 (main, Mar  2 2022, 04:23:34) [GCC 10.2.1 20210110]
  jinja version = 3.1.2
  libyaml = True

$ pip freeze
absl-py==1.1.0
ansible==6.1.0
ansible-core==2.13.1
bcrypt==3.2.2
capirca==2.0.6
cffi==1.15.1
cfgv==3.3.1
click==8.1.3
cryptography==37.0.4
distlib==0.3.4
filelock==3.7.1
ghp-import==2.1.0
identify==2.5.1
importlib-metadata==4.12.0
ipaddress==1.0.23
Jinja2==3.1.2
junos-eznc==2.6.4
jxmlease==1.0.3
lxml==4.9.1
Markdown==3.3.7
MarkupSafe==2.1.1
mergedeep==1.3.4
mkdocs==1.3.0
mkdocs-material==8.3.9
mkdocs-material-extensions==1.0.3
mock==4.0.3
ncclient==0.6.13
netaddr==0.8.0
nodeenv==1.7.0
packaging==21.3
paramiko==2.11.0
pathspec==0.9.0
platformdirs==2.5.2
ply==3.11
pre-commit==2.20.0
pycparser==2.21
Pygments==2.12.0
pymdown-extensions==9.5
PyNaCl==1.5.0
pyparsing==3.0.9
pyserial==3.5
python-dateutil==2.8.2
PyYAML==6.0
pyyaml_env_tag==0.1
resolvelib==0.8.1
scp==0.14.4
six==1.16.0
toml==0.10.2
transitions==0.8.11
virtualenv==20.15.1
watchdog==2.1.9
xmltodict==0.13.0
yamllint==1.27.1
yamlordereddictloader==0.4.0
zipp==3.8.1

Collection     Version
-------------- -------
[...]

Issue #582, opened 2022-01-12
RpcTimeoutError commit-configuration


csk06:
Issue Type
------

 - Bug Report

Module Name
------

juniper.device collection and Python libraries version

```
ansible [core 2.12.1]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/ubuntu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/ubuntu/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.8.10 (default, Nov 26 2021, 20:14:08) [GCC 9.3.0]
  jinja version = 2.10.1
  libyaml = True

junos-eznc==2.6.3
```

OS / Environment
------

Summary
------

Hi, I am trying to configure a lab device with following commands, but even for simple hostname changing, half of the times im getting the error below. As i understand, that "commit-configuration" is coming from python library, and there is variable to change that value something different from 10, but on ansible i guess there is no variable to change it. 

Steps to reproduce
------

```
- name: "Configure Device" 
  hosts: junos
  connection: local
  timeout: 50 
  gather_facts: no

  tasks:
    - name: "Configure op script"
      juniper.device.config:
        config_mode: "private"
        check_commit_wait: 20
        load: "set"
        src: "test.txt" 
      register: response
    - name: "Print the config changes"
      debug:
        var: response
```

Expected results

```

```
Actual results
------

```
fatal: [172.31.30.158]: FAILED! => {"changed": false, "msg": "Failure checking the configuraton: RpcTimeoutError(host: 172.31.30.158, cmd: commit-configuration, timeout: 10)"}
```

ric79:
Same error for me

Failure committing the configuraton: RpcTimeoutError(host: xx.xx.xx.xx, cmd: commit-configuration, timeout: 10)

Did you solve it?

```
    - juniper.device.config:
        load: "merge"
        lines:
           - ....
        comment: "..."
        commit: true
        commit_empty_changes: true
        config_mode: "exclusive"
      vars:
        ansible_connection: local

```

ric79:
Workaround:

https://github.com/Juniper/ansible-junos-stdlib/blob/master/ansible_collections/juniper/device/plugins/module_utils/juniper_junos_common.py#L1529

```
        self.logger.debug("Committing the configuration.")
        try:
[...]

PR #591, opened 2022-07-22
Improve handling of complex RPC's

TheMysteriousX:
These changes work for me to resolve the issues I noted in #589 - I believe them to be safe and not cause any changes in behaviour (mostly because it doesn't appear the current behaviour works) but I don't have an extensive library of tasks to check these against as I'm working on a brand new deployment. But you may want to rework them, as I'm not familiar with ansible's best practices for module development.

This change does three things:
* Port usage of _check_type_dict to support Ansible 4+
I've tried to maintain compatibility with Ansible 2.9 by adding a fallback.

* Changes the type of attrs, kwargs, and filter to list, avoiding round-tripping them from list to string to list. This is a problem, because safe_eval will not process lists with certain keywords that are common in JunOS configs.
I believe this is a safe change, as if a user has e.g. "kwarg: 'foo'" Ansible will transparently convert it to "['foo']".

* Allows an RPC to have nested configuration - this is required for e.g. validate, which needs to generate:
```
<rpc>
  <validate>
    <candidate/>
  </validate>
</rpc>
```
This is done recursively.

### DIFF
--- a/ansible_collections/juniper/device/plugins/module_utils/juniper_junos_common.py
+++ b/ansible_collections/juniper/device/plugins/module_utils/juniper_junos_common.py
@@ -1518,7 +1518,7 @@ class JuniperJunosModule(AnsibleModule):
                                (str(ex)))
 
     def commit_configuration(self, ignore_warning=None, comment=None,
-                             confirmed=None, timeout=30, full=False):
+                             confirmed=None, full=False):
         """Commit the candidate configuration.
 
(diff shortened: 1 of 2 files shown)


TheMysteriousX:
Also includes a fix for #582 now

dineshbaburam91:
As per  @chidanandpujar  comment, it is working without your changes.

PR #634, opened 2023-11-23
feat: Enable using only the SSH agent for private keys if asked.

gaima8:
jnpr (py-junos-eznc) will only ask the SSH agent for keys in certain circumstances, this allows you to override that and for it to ask an agent.

A PR for Juniper/py-junos-eznc will be made too.

### DIFF
--- a/ansible_collections/juniper/device/plugins/connection/pyez.py
+++ b/ansible_collections/juniper/device/plugins/connection/pyez.py
@@ -113,7 +113,7 @@ options:
     env:
     - name: ANSIBLE_PYEZ_CONSOLE
     vars:
-    - name: ansible_pyez_console    
+    - name: ansible_pyez_console
   private_key_file:
     description:
     - The private SSH key or certificate file used to authenticate to the remote device
@@ -202,7 +202,17 @@ options:
     vars:
     - name: ansible_pyez_ssh_config
     - name: ssh_config
-    
+  force_agent:
+    description:
+    - This flag will request that only the SSH agent (if found) will be queried for keys.
+    ini:
+    - section: defaults
+      key: force_agent
+    env:
+    - name: ANSIBLE_FORCE_AGENT
+    vars:
+    - name: ansible_force_agent
+    - name: force_agent
 """
 import pickle
 
@@ -381,6 +391,7 @@ class Connection(NetworkConnectionBase):
         connect_args['ssh_private_key_file'] = self.get_option('private_key_file')
         connect_args['ssh_config'] = self.get_option('pyez_ssh_config')
         connect_args['timeout'] = self.get_option('persistent_connect_timeout')
+        connect_args['allow_agent'] = self.get_option('force_agent')
         try:
             log_connect_args = dict(connect_args)
             log_connect_args["passwd"] = "NOT_LOGGING_PARAMETER"


gaima8:
Probably superseded by https://github.com/Juniper/py-junos-eznc/pull/1285

chidanandpujar:
Hi @gaima8 
     Thanks ,
     Please share the test playbooks and execution results for review .
    https://github.com/Juniper/ansible-junos-stdlib/tree/master/tests

Thanks

gaima8:
```
$ grep -o ^IdentityFile ~/.ssh/config 
IdentityFile
$ ansible-playbook ....whatever....
...Unable to make a PyEZ connection: ConnectAuthError(........
```

The value of `IdentityFile` is an encrypted private key file.

In junos-eznc where `self._conf_ssh_private_key_file` is set by `class _Connection._sshconf_lkup`, causing `self._ssh_private_key_file` to be set by `class Device.__init__`, resulting in `allow_agent` being false causing `netconf_ssh.connect` to not use the agent.

All this change does is allow you to set force_agent, with https://github.com/Juniper/py-junos-eznc/pull/1284 this just forces junos-eznc to set allow_agent to true in the call to ncclient.
By setting `force_agent` to True the playbook works.


As ncclient itself defaults to having allow_agent=True then https://github.com/Juniper/py-junos-eznc/pull/1285 is probably a better fix as that simply stops junos-eznc from ever setting allow_agent to any other value.
However ssh does actually allow you to not use the ssh-agent (enable IdentitiesOnly to disable using the agent) so it is arguable whether junos-eznc should continue to allow the agent to be disabled. This PR gives you that same option, keeping the current behaviour intact.


      IdentitiesOnly
             Specifies that ssh(1) should only use the configured authentication identity and certificate files (either the default files, or those explicitly configured in the ssh_config files or
             passed on the ssh(1) command-line), even if ssh-agent(1) or a PKCS11Provider or SecurityKeyProvider offers more identities.  The argument to this keyword must be yes or no (the de‐
             fault).  This option is intended for situations where ssh-agent offers many different identities.

chidanandpujar:
> ```
> $ grep -o ^IdentityFile ~/.ssh/config 
> IdentityFile
> $ ansible-playbook ....whatever....
> ...Unable to make a PyEZ connection: ConnectAuthError(........
> ```
> 
> The value of `IdentityFile` is an encrypted private key file.
> 
> In junos-eznc where `self._conf_ssh_private_key_file` is set by `class _Connection._sshconf_lkup`, causing `self._ssh_private_key_file` to be set by `class Device.__init__`, resulting in `allow_agent` being false causing `netconf_ssh.connect` to not use the agent.
> 
> All this change does is allow you to set force_agent, with [Juniper/py-junos-eznc#1284](https://github.com/Juniper/py-junos-eznc/pull/1284) this just forces junos-eznc to set allow_agent to true in the call to ncclient. By setting `force_agent` to True the playbook works.
> 
> As ncclient itself defaults to having allow_agent=True then [Juniper/py-junos-eznc#1285](https://github.com/Juniper/py-junos-eznc/pull/1285) is probably a better fix as that simply stops junos-eznc from ever setting allow_agent to any other value. However ssh does actually allow you to not use the ssh-agent (enable IdentitiesOnly to disable using the agent) so it is arguable whether junos-eznc should continue to allow the agent to be disabled. This PR gives you that same option, keeping the current behaviour intact.
> 
> ```
>   IdentitiesOnly
>          Specifies that ssh(1) should only use the configured authentication identity and certificate files (either the default files, or those explicitly configured in the ssh_config files or
>          passed on the ssh(1) command-line), even if ssh-agent(1) or a PKCS11Provider or SecurityKeyProvider offers more identities.  The argument to this keyword must be yes or no (the de‐
>          fault).  This option is intended for situations where ssh-agent offers many different identities.
> ```

Hi @gaima8 
     Thanks for sharing the info .
      I will go through it and update you.

Thanks

Issue #607, opened 2023-01-26
Timeout value for rpc timeout not used


T0nip:
Issue Type
------

 - Bug Report

Module Name
------

juniper.device.config

juniper.device collection and Python libraries version

```
ansible --version
ansible [core 2.14.1]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```
```
pip freeze
ansible==7.1.0
ansible-core==2.14.1
ansible-pylibssh==1.1.0
bcrypt==4.0.1
cffi==1.15.1
cryptography==39.0.0
Jinja2==3.1.2
junos-eznc==2.6.6
jxmlease==1.0.3
lxml==4.9.2
MarkupSafe==2.1.2
ncclient==0.6.13
netaddr==0.8.0
packaging==23.0
paramiko==3.0.0
pycparser==2.21
PyNaCl==1.5.0
pyparsing==3.0.9
pyserial==3.5
PyYAML==6.0
resolvelib==0.8.1
scp==0.14.4
six==1.16.0
supervisor==4.2.2
transitions==0.9.0
xmltodict==0.13.0
yamlordereddictloader==0.4.0
```

OS / Environment
------

Juniper srx300 21.2R3-S2.9
Summary
------

The provided value for timeout isn't used in modules, so the playbook fails while commiting changes.

Steps to reproduce
------

Commit changes that take longer than 30s.

```yaml
        - name: install new config
          config:
            host: "{{ inventory_hostname }}"
            port: "{{ sshport }}"
            user: "{{ username }}"
            passwd: "{{ password }}"
            load: 'set'
            src: "{{ config_path }}{{ config_file }}"
            logfile: "junos.log"
            confirmed: 20
            timeout: 120
```

Expected results

```
timeout value should be used, so setting 120s should adjust the timeout used, to 120s.
```
Actual results
------

playbook stops with timeout error after 30s

```

```

IvanBCode:
the same issue
is there any workaround?

m0msg:
[...]

PR #641, opened 2024-02-13
implement rpc timeout 

apurvaraghu:
fix issue #607

### DIFF
--- a/ansible_collections/juniper/device/changelogs/config.yaml
+++ b/ansible_collections/juniper/device/changelogs/config.yaml
@@ -1,0 +1,32 @@
+changelog_filename_template: ../CHANGELOG.rst
+changelog_filename_version_depth: 0
+changes_file: changelog.yaml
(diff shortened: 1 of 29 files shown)


n3wtype:
I just investigated this issue. It seems that timeout behaviour of commit_confirmed got broken with this commit: https://github.com/Juniper/ansible-junos-stdlib/commit/0cdae0e4d0274c5826da5ba74dec0ecbf7c1e999 .

It introduced timeout param for commit_confirmed with default value of 30. 

However, the only place where commit_configuration is called is https://github.com/Juniper/ansible-junos-stdlib/blob/9e39e0023f25ffa4e4d430d5905e60299018ad9a/ansible_collections/juniper/device/plugins/modules/config.py#L1229 and timeout parameter is not passed here, so default value is always used. 

I think more proper solution would be to pass "timeout" parameter to this call if it was passed to module.

PR #646, opened 2024-03-27
Fixed functional testcases for ping and pmutd playbook

dineshbaburam91:
Issue:
Normalize value not passed correctly in PyEZ connection rpc.ping() API


chidanandpujar (approved):
looks good

PR #665, opened 2024-05-16
Fix for the issue #607

chidanandpujar:
added timeout parameter and passed it to commit_configuration
example:
    - name: "Load configuration from a local file and commit"
      juniper.device.config:
        load: "merge"
        src: "junos_jsnapy/junos.conf"
        confirmed: 20
        timeout: 120


dineshbaburam91 (approved):
Looks good

PR #677, opened 2024-07-12
Fixed rpd module related issue

dineshbaburam91:
When performing an RPC get with an RPC that is not implemented for a platform, it fails when using a PyEZ connection. With a local connection, the response is “Unable to execute the RPC” and it continues with the next RPC get. PyEZ behavior should be the same as with the local connection.


chidanandpujar (approved):
Looks good

Issue #690, opened 2024-08-28
juniper.device.facts with config_format: "json" with persistent connection juniper.device.pyez fails with ConnectionError: Type 'dict' cannot be serialized.\n"


chidanandpujar:
Issue Type
------
 - Bug Report

Module Name
------
juniper.device.facts

juniper.device collection and Python libraries version
```
(venv) root@masterhost:~/ansible_doc_test/ansible-junos-stdlib/tests# ansible --version
ansible [core 2.17.3]
  config file = /root/ansible_doc_test/ansible-junos-stdlib/tests/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/ansible_doc_test/venv/lib/python3.12/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /root/ansible_doc_test/venv/bin/ansible
  python version = 3.12.0 (main, Aug 27 2024, 22:17:06) [GCC 9.3.0] (/root/ansible_doc_test/venv/bin/python3.12)
  jinja version = 3.1.4
  libyaml = True
(venv) root@masterhost:~/ansible_doc_test/ansible-junos-stdlib/tests# pip freeze
ansible==10.3.0
ansible-core==2.17.3
bcrypt==4.2.0
cffi==1.17.0
colorama==0.4.6
configparser==7.1.0
cryptography==43.0.0
future==1.0.0
icdiff==1.9.1
Jinja2==3.1.4
jsnapy==1.3.8
junos-eznc @ git+https://github.com/Juniper/py-junos-eznc.git@a64698b280fb717730dddf75a2d4d8cacb2e4775
jxmlease==1.0.3
lxml==5.3.0
MarkupSafe==2.1.5
ncclient==0.6.15
packaging==24.1
paramiko==3.4.1
pycparser==2.22
PyNaCl==1.5.0
pyparsing==3.1.4
pyserial==3.5
PyYAML==6.0.2
resolvelib==1.0.1
scp==0.15.0
setuptools==74.0.0
six==1.16.0
transitions==0.9.2
xmltodict==0.13.0
yamlordereddictloader==0.4.2
(venv) root@masterhost:~/ansible_doc_test/ansible-junos-stdlib/tests# 


```

Summary
------
[...]

PR #691, opened 2024-08-30
Fix for issue #690

chidanandpujar:
Added code to check if response is in json format then return the response as it is.


chidanandpujar:
fix is handled in pull request https://github.com/Juniper/ansible-junos-stdlib/pull/703

PR #695, opened 2024-10-01
Fix version extraction from filename for ex2300 firmware (fixes #694)

### LINKED ISSUE #694: junos.device.software not parsing version from filename correctly for EX2300
Issue Type
------
 - Bug Report

Module Name
------
junos.device.software

juniper.device collection and Python libraries version

```
juniper.device 1.0.6

ansible [core 2.16.11]
  python version = 3.12.6 (main, Sep  9 2024, 00:00:00) [GCC 13.3.1 20240522 (Red Hat 13.3.1-1)] (/usr/bin/python3)
  jinja version = 3.1.4
  libyaml = True
```

OS / Environment
------
Junos 23.4

Summary
------
junos.device.software expects the filename to start with one of these prefixes to actually parse the version string:
```
    JUNOS_PACKAGE_PREFIXES = ['jbundle', 'jinstall', 'junos-install',
                              'junos-srx', 'junos-vmhost-install', 'junos-vrr',
                              'vmx-bundle']
```

However, the EX2300 recommended release 23.4 package file is named `junos-arm-32-23.4R2.13.tgz`. Therefore the version is not extracted, causing firmware to be installed when it is already the current version.

Steps to reproduce
------
Download firmware for these switches from: https://support.juniper.net/support/downloads/?p=ex2300

and try to update to this firmware on a switch that already has the relevant firmware installed. (without explicitly specifying version parameter)

Yannik:
This fixes #694.

### DIFF
--- a/ansible_collections/juniper/device/plugins/modules/software.py
+++ b/ansible_collections/juniper/device/plugins/modules/software.py
@@ -417,7 +417,7 @@ def parse_version_from_filename(filename):
     # Known prefixes for filenames which contain Junos software packages.
     JUNOS_PACKAGE_PREFIXES = ['jbundle', 'jinstall', 'junos-install',
                               'junos-srx', 'junos-vmhost-install', 'junos-vrr',
-                              'vmx-bundle']
+                              'vmx-bundle', 'junos-arm']
     for prefix in JUNOS_PACKAGE_PREFIXES:
         if filename.startswith(prefix):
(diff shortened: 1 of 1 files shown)


chidanandpujar:
Hi @Yannik 
    Thanks for submitting the pull request.
    we will review the fix and update.

Thanks
Chidanand

chidanandpujar (approved):
Looks good.

PR #735, opened 2025-04-22
Enhanced file_copy module

dineshbaburam91:
### Features Added

1. A new option, "protocol," has been added to allow users to choose the file transfer method: SCP or FTP. This is an optional parameter, with SCP set as the default.

2. A new option, "checksum," has been introduced to enable or disable validation using a boolean value of True or False. This is also an optional parameter, with True set as the default. When enabled, it validates the MD5 checksum algorithm before and after the file transfer. If set to False, no validation will be performed.

3. A new optional parameter, "transfer-file," allows you to specify the name of the file to be transferred. For example, if you want to copy a file named "XYZ" to a remote destination with a different name, "ABC," you can specify "ABC" in the transfer-file parameter. By default, the parameter will use the same name as the file being transferred.

### Sample Playbook

```
    - name: "File copy put by ftp with transfer filename parameter"
      juniper.device.file_copy:
        protocol: ftp
        local_dir: /var/tmp
        remote_dir: /var/log
        file: transfer_log.txt
        transfer_filename: remote_log.txt
        checksum: false
        action: put
      register: test1
      ignore_errors: true
```

### DIFF
--- a/ansible_collections/juniper/device/plugins/connection/pyez.py
+++ b/ansible_collections/juniper/device/plugins/connection/pyez.py
@@ -247,6 +247,13 @@ except ImportError:
     HAS_PYEZ_SCP = False
 
 try:
+    from jnpr.junos.utils.ftp import FTP
+
+    HAS_PYEZ_FTP = True
+except ImportError:
+    HAS_PYEZ_FTP = False
+
+try:
     import jnpr.junos.utils.config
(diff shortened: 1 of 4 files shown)


chidanandpujar (changes requested):
please provide the FT and UT pass logs

PR #777, opened 2025-10-05
remove empty key action to fix the nonetype warning

Ruchip16:


chidanandpujar:
> hi @chidanandpujar why are we doing the changes in [ansible-collections/junipernetworks.junos#574](https://github.com/ansible-collections/junipernetworks.junos/pull/574), i dont think its needed, only my suggested change to fix those warnings is needed here : #777

Hi @Ruchip16 ,
    This was already discussed with Rohit, to support redirection to juniper.device collection path for Redhat ansible collection users,  so that existing Redhat  customer playbooks continue to work with redirection support.
 

Thanks

PR #790, opened 2025-11-19
Update examples section with YAML truthy values

oraNod:
ansible-lint warns that truthy values should be true / false because the examples section contains YAML strings.

Observed this in the import logs for the collection upload to Automation Hub at: https://console.redhat.com/ansible/automation-hub/my-imports/?namespace=juniper


chidanandpujar:
fixed from another pull request, so closing this pull request

PR #805, opened 2026-02-16
Add deprecation warnings for junipernetworks.junos ->  juniper.device redirects

rohitthakur2590:
All plugin and module redirects from junipernetworks.junos to juniper.device now emit a deprecation warning. Redirects stay in place until 2027-03-01; after that they may be removed.
Changes:
meta/runtime.yml: added deprecation (with removal_date and warning_text) to every redirect.
Changelog fragment under Deprecated Features and CHANGELOG.rst updated.
User impact: Existing playbooks keep working. Users will see a warning and should migrate to juniper.device FQCNs (e.g. juniper.device.junos_config) before the removal date.

image


chidanandpujar (approved):
looks good

PR #838, opened 2026-07-14
safe_eval is replaced with ast

chidanandpujar:
self.safe_eval(string_val)  is replace with ast.literal_eval(string_val)


dineshbaburam91 (approved):
Looks good. It is must needed changes for latest ansible core version

PR #876, opened 2026-08-26
Enable & Disable logical interface unit

dineshbaburam91:
## Description

Enable & Disable logical interface unit in junos_interface module

Closes #

---

## Collection Namespace

- **Collection namespace:** 
- **Collection version (bumped to):** 
- **Module(s) / plugin(s) changed:** 

---

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] Feature / new module or plugin (non-breaking change that adds functionality)
- [ ] Enhancement (improvement to existing module or plugin behaviour)
- [ ] Breaking change (fix or feature that would cause existing playbooks / roles to behave differently)
- [ ] Task / Chore (refactor, dependency update, CI, docs, etc.)

---

## Fix Details

Enabled & Disabled logical interface unit in junos_interface module adding enable and disable parameter support

---

## Versions Tested

| Item | Version |
|------|---------|
| Collection namespace & version | juniper.device 2.9.4 |
| Ansible / ansible-core | ansible-core latest |
| Python | 3.13.0 |
| OS | mac |
| Junos version | 26.4 |
| Junos platform | ptx |
| junos-eznc | 2.7.1 |
| ncclient | 0.6.15  |

---

## Sanity Test Results

- [X] Sanity tests pass with no new errors

```bash
ansible-test sanity
```

<details>
<summary>Sanity test output</summary>

```
<paste sanity test output here>
```

</details>

---

## Unit Test Results

- [ ] New or updated unit tests added under `tests/unit/`
- [X] All unit tests pass locally

```bash
python3.12 -m pytest
```

<details>
<summary>Unit test output</summary>

```
<paste unit test output here>
```

</details>

- [ ] Code coverage has not decreased

---

## Integration Test Results

- [X] Integration tests reviewed for impact
- [ ] Affected integration tests pass (requires a live Junos device or vMX)

```bash
ansible-test network-integration --python 3.12 --inventory /root/pyez_ansible_release_validation1/ansible-junos-stdlib/ansible_collections/juniper/device/tests/integration/inventory.networking
```

<details>
<summary>Integration test output</summary>

```
<paste integration test output here>
```

</details>

---

### Code Quality

- [X] Code follows the project's style guidelines
- [ ] `ansible-lint` passes with no new errors

```bash
ansible-lint
```

- [ ] `flake8` / `ruff` reports no new errors (for Python files)

```bash
flake8 plugins/
# or
ruff check plugins/
```

### Documentation

- [ ] Module documentation (DOCUMENTATION, EXAMPLES, RETURN blocks) updated if parameters changed
- [ ] `CHANGELOG.rst` / `changelogs/fragments/` entry added
- [ ] `galaxy.yml` version bumped if releasing

### General

- [ ] No hardcoded credentials, IP addresses, or sensitive data introduced
- [ ] `requirements.txt` updated if new Python packages are required
- [ ] Backward-compatible: existing playbooks will not break

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -145,11 +145,14 @@ jobs:
           # 2. Install ncclient (Required to fix the 'new_ele' bug in netcommon)
           python -m pip install ncclient
 
-          # 3. Install juniper.device under the Junos collection's local search root.
-          # tox runs from that directory with ANSIBLE_COLLECTIONS_PATH=., so installing
-          # to the default user path would allow a stale collection to be selected.
+          # 3. Install the local juniper.device so it satisfies the
+          # "juniper.device: >=2.0.0" dependency instead of the stale Galaxy-published
+          # collection. Install it both under the Junos collection's local search root
+          # (tox runs from there with ANSIBLE_COLLECTIONS_PATH=.) and the default path
+          # (used by ade dependency resolution), so the local code always wins.
           ansible-galaxy collection install ansible_collections/juniper/device/ --force \
             -p ansible_collections/junipernetworks/junos
+          ansible-galaxy collection install ansible_collections/juniper/device/ --force
       - name: "Install tox-ansible, includes tox"
         run: python -m pip install tox-ansible
       - name: "Check for tox-ansible.ini file, else add default"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,8 +70,6 @@ jobs:
             ansible: "2.19"
           - python: "3.12"
             ansible: "2.19"
-          - python: "3.13"
-            ansible: "2.19"
     steps:
       - uses: actions/checkout@v4
         with:
@@ -100,7 +98,7 @@ jobs:
       - name: Run tox sanity tests
         if: matrix.collection == 'juniper/device'
         working-directory: ansible_collections/${{ matrix.collection }}
-        run: ansible-test sanity --python 3.13
+        run: ansible-test sanity --python 3.12
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       - name: Run tox sanity tests
@@ -124,8 +122,6 @@ jobs:
           - python: "3.11"
             ansible: "2.19"
           - python: "3.12"
-            ansible: "2.19"
-          - python: "3.13"
             ansible: "2.19"
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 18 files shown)


chidanandpujar (commented):
Please check following comments, any UT needs to be covered


test_junos_interfaces.py:114-136 — test_junos_interfaces_logical_unit_enabled_states

This test passes commands into execute_module, but execute_module/load_fixtures never compares it against result["rendered"] — it only uses commands to build the fixture. As written, the test passes even if the rendered XML is wrong or omits <enable/>/<disable/>. Please add an explicit assertion, e.g. self.assertEqual(result["rendered"], commands[0]).

interfaces.py:176-179 — facts parsing, single-unit branch

Can you clarify what produces the "undocumented" key here? It's not obvious from the diff what Junos/xmltodict scenario this branch is handling, and it currently has no test coverage. A short comment explaining when this key shows up would help future maintainers.

interfaces.py:192-195 — facts parsing, multi-unit branch

Same "undocumented" handling is duplicated here from the single-unit branch above. Consider extracting the shared logic (name/description/vlan_id/enabled extraction) into a helper to avoid maintaining two copies.

interfaces.py:372-385 — _state_deleted, per-unit

This mirrors the have_cfg.get("enabled") is False check above for the interface-level disable delete, but for units it deletes whichever tag is present (enable or disable) rather than only handling the disabled case. Worth a short comment noting that distinction is intentional, since it's easy to misread as duplicate/redundant logic.

No test exercises merged/replaced/deleted with unit-level enabled, nor gathered/parsed facts parsing for <unit><enable/></unit> and <unit><disable/></unit>. Only the rendered state is covered. Given this changes both the config renderer and the facts parser, please add coverage for at least one round-trip (merge + idempotent) and one facts-parsing test with a unit-level enable/disable fixture.

chidanandpujar (commented):
Please check and address 

High: Enabling a logical unit does not clear an existing disable state.
At config/interfaces/interfaces.py line 241, enabled=True renders <enable/>. A Junos unit that currently contains <disable/> needs that node deleted; adding <enable/> does not remove the existing disable configuration and may be rejected by Junos. Please model enabled=True as deletion/absence of <disable/>, and enabled=False as <disable/>. Add a test for transitioning an existing disabled unit to enabled.

High: The new interface tests expect output that their input cannot produce.
In test_junos_interfaces.py line 328, and the new logical-unit tests around line 130, the expected XML includes an interface-level <enable/>, but the supplied config does not set enabled on the parent interface. _state_merged() only emits that element when config.get("enabled") is not None, so these tests should fail. Either set enabled=True in the test input or remove the unexpected parent <enable/> from the expected XML.

Additional test gap

The change needs coverage for state: deleted and state: replaced when logical units transition between enabled and disabled states. The current tests only cover rendering, merging, and parsing.

chidanandpujar (approved):

